PR #15, opened 2017-03-07
Using Kratos Model

rubenzorrilla:
Instead of a standard dictionary, now the Kratos Model() is used.


rubenzorrilla:
@lorenzogracia @ipouplana I've tried the changes for both the fluid and structural apps but a dam app test remains to be done. I'll wait until your test approval to merge the changes.

lorenzogracia:
I agree with the changes. I approve your pull request. (Pure cream)

Issue #59, opened 2017-04-06
Outlet process


jginternational:
Unify outlet processes into a single one

Must have the classical parameters plus:
- hydrostatic_outlet
- h_top

jginternational:
Ready!
https://github.com/KratosMultiphysics/Kratos/pull/259

PR #61, opened 2017-04-06
Update outlet process

jginternational:
As said in #59 and https://github.com/KratosMultiphysics/Kratos/pull/259 This pull request creates the unified Outlet Condition

Thanks to this contribution, there are some kind of parameters in processes that accept dependency


rubenzorrilla:
@jginternational I've tried a case with and without body force (with and without h_top) and it works with no problems. However, an error pop-up appears when the problem type tries to load the predefined cylinder example button. Is this expected?

jginternational:
This happens because you opened the master branch before this one, so GiD remembers that tcl. Should be   fine if you checkout this branch before loading GiD

rubenzorrilla:
@jginternational  I thought it was something like that. I was just to be sure before merging. Thanks!

Issue #66, opened 2017-04-07
Dam examples


jginternational:
Follow the steps:

- [x] Create the files and buttons
- [x] Implement the geometry creation
- [x] Implement the Group assign
- [x] Implement the Mesh sizes  (optional)
- [x] Implement the General parameters and conditions assign

PR #73, opened 2017-04-08
Example finished

lorenzogracia:
#66


jginternational:
Bravo! Is it only for 2D?

lorenzogracia:
Yes, the geometry for a 3D dam model is much more complicated, so to understand how must be assigned all conditions is enough. In the future an example with a complex geometry can be implemented.

jginternational:
So, i'll hide the example button in 3D

Issue #70, opened 2017-04-07
DAM Change deprecated process parameters from XYZ to verctor


jginternational:
DAM application processes are using a deprecated function `[CheckDimension 3D]` to show/hide the Z component in 2D models

The new way to implement this is by using vector and vector type
Example:
```xml
<inputs>
    <parameter n="is_fixed_X" pn="Imposed X" type="bool" v="1"/>
    <parameter n="is_fixed_Y" pn="Imposed Y" type="bool" v="1"/>
    <parameter n="is_fixed_Z" pn="Imposed Z" type="bool" v="1" state="[CheckDimension 3D]"/>
</inputs>
```
is now:
```xml
<inputs>
    <parameter n="is_fixed" pn="Imposed" type="vector" vectorType="bool" v="1,1,1"/>
</inputs>
```

lorenzogracia:
@jginternational, the changes in type=vector are already committed. The point is that I also use that process for the type=table, are there something similar for tables??

jginternational:
At this point, boolean and doble 3 dimensional parameters are written in a vector way, as we can see it in is_fixed and value
```json
"Parameters"    : {
            "mesh_id"         : 0,
            "model_part_name" : "DISPLACEMENT_Displacement",
            "variable_name"   : "DISPLACEMENT",
            "is_fixed"        : [true,true,true],
            "value"           : [0.0,0.0,0.0],
            "Value_X_Table"   : 0,
            "Value_Y_Table"   : 0,
            "Value_Z_Table"   : 0
        }
```
3 dimensional tables must be updated to be written the same way.
The result will be like this:
```json
"Parameters"    : {
            "mesh_id"         : 0,
            "model_part_name" : "DISPLACEMENT_Displacement",
            "variable_name"   : "DISPLACEMENT",
            "is_fixed"        : [true,true,true],
            "value"           : [0.0,0.0,0.0],
            "Value_Table"   : [0,0,0]
        }
```
I'll update the way to implement this for everybody, but check your python process

lorenzogracia:
Ok @jginternational. When it will available, please, let me know. I am going to close this issue.

PR #75, opened 2017-04-10
#70 #72 Table parameters can be three dimensional (X Y Z)

jginternational:
Fixes the issues of the deprecated state: CheckDimension on tables


lorenzogracia:
I have been checking it and everything works fine, I also have updated my process in Kratos.

Issue #79, opened 2017-04-12
Generic submodelpart from group


jginternational:
As said in https://github.com/KratosMultiphysics/Kratos/issues/278

There is a proposal that needs a  `<condition>` to write a Group as submodelpart.
something like:
```json
<condition n="GenericModelParts" pn="Add SubModelPart" un="ExtraModelParts" ov="point,line,surface,volume" state="[ShowInMode Developer]">
    <value n="WriteElements" pn="Write elements" v="False" values="True,False"/>
    <value n="WriteConditions" pn="Write conditions" v="True" values="True,False"/>
</condition>
```
The result is:
![image](https://cloud.githubusercontent.com/assets/5918085/24953516/e58dd5d8-1f7b-11e7-80ef-b9b5a8d30bb5.png)
**Note:** It's important to select the entity type even if the group was previously created

We'll need to modify the writing process, to add this new feature

adityaghantasala:
Hello @jginternational  ... I have a question regarding the usage of this problem type. I downloaded this branch and had a look at it. I saw that the generic submodelpart from group works for Structural as mentioned. 

Is it also possible to do it for a fluid problem (2D and 3D) ?? If not is there a plan to do that ?

jginternational:
The idea is to have it in all apps (in development mode) soon

adityaghantasala:
@jginternational  that would be awesome .. thank you!

Issue #80, opened 2017-04-13
How to write Conditions


jginternational:
In the issue https://github.com/KratosMultiphysics/Kratos/issues/278 @RiccardoRossi , @maceligueta , @philbucher and me were talking about How should the problemtype write the conditions:

I've written this, trying to clearify some ideas (forgive me for the extension):

From the problemtype side, the most efficient way is to **print the GiD element id**. 
Let's call this **the A way** of writing the conditions.
That is done using the customLib C++ lib, so it is faster and allows to handle larger number of elements than the current way of printing the conditions' block, in TCL.

**The B way** is do what we are doing now, each app implements it's way of printing/storing the conditions (or uses the common functions to do so). At this point, the pseudo-code looks like this:

In the Begin Conditions sections:
```
foreach condition in conditions: //(in the tree)
    foreach group in GetGroupsOf(condition)
        print "Begin Condition _ConditionName_"
        foreach element in GetElementsOf(group)
            nodes = GetNodesOf(element)
            print "i 0 nodes"
            ++i
        print "End Condition"
        Store(condition,group, initial_id, end_id)
```
Each condition-group stores its interval of ids, to print them later in the submodelpart.

In the submodelparts sections:
```
foreach condition-group in GetStoredConditionsGroups // previously stored
    Print "Begin SubModelPart _condition-group_
    Print "Begin nodes"
    foreach node in GetNodes(group)
        print node
    Print "End nodes"
    Print "Begin conditions"
    i = condition-group.getInitial
    while i < condition-group.getFinal
        print i
        ++i
    print "End conditions"
    print "End subModelPart"
```

Let's compare

**THE A WAY**:
pros: 
- Increased performance (C++ writing)
- Easier implementation (we do not need to keep control of assigned ids)

cons:
- Points are meshed as nodes, not as '1 noded elements', so it's ids are assigned as nodal id (We'll see it in an example). Dr. Abel Coll (manager of the GiD dept.) has explained me that he can change something in the GiD tcl API to get nodes meshed as elements.
- Conditions id will not be consecutive
- As @maceligueta said, it will be impossible to get different id for the same entity, if you need so.

**THE B WAY**:
pros: 
- It is already implemented
- Conditions on nodes have a valid id assigned, without telling GiD to change it's meshing criteria

cons:
- Lower performance (TCL writing)
- Lower amount of elements (TCL can store lower information in a variable, remember the BCN model problems)
- Difficult implementation
[...]

PR #84, opened 2017-04-21
Feature/add generic submodelparts

jginternational:
This completes #79 and (https://github.com/KratosMultiphysics/Kratos/issues/278)

The new feature adds to the Structural Application a section (only in developer mode) to create Submodelparts with no process assigned. This means that the submodelpart will be added to de MDPA but will have no effects


philbucher:
Nice work, this is exactly what I thought of!
I tested it, it does the correct thing. I will test it in more detail in the future and let you know in case sth is not working.

I have three Questions: 
1. It only works with the developer version of GiD for me (and being in developer mode) In the normal version of GiD (13.0.2) it says that GiD 13.1.1d or later is required. Is that intended?
2. When I run kratos to get the mdpa file it always says that I the geometry has changed and I should probably remesh, even though the geometry hasn't changed and I saved / remeshed the model. Am I doing sth wrong here?
3. Will you leave it in the Structural Problemtype or create a separate one? Or integrate it in every problemtype (esp. for fluid I think it would be useful)?
4. Nodes have to always be written to the submodelpart, right?

I also have a comment/suggestion for possible further improvement: Would it be possible to change the type of the elements/conditions with an arbitrary string? I mean that if I e.g. implement a new element I could exchange the name in GiD and wouldn't have to got through the mdpa file each time.

Thanks a lot for your effort!

jginternational:
Hi @philbucher ! I'll answer point by point.

1. In fact, the GiD team has noticed (and fixed) a bug that affected many problemtypes for Windows. The bug is fixed in the GiD 13.1.6d (not released yet but I'm testing it and seem to be fixed.
2. Same than 1, I'm going to ask the team if I could give you an installer of the new version.
3. My idea is to integrate it in the main applications, Solid, Fluid, Structural, Pfem....
4. Yes, because if elements and conditions connectivities are needed, I can not control what nodes have been already written, so the best policy is to write any single submodelpart as a standalone part, even if we have to write a previously written node

About the further improvement, I'll need to change some rules but could be implemented easily, 

Let me discuss this points with @RiccardoRossi and this branch should be merged soon this week

philbucher:
Hi @jginternational,
Thx for your answers.
I just discussed with @dbaumgaertner and he also thinks that my suggestion with replacing the elements/conditions in GiD would be nice.

Also it is important to us that we don't have to set up an entire analysis to write the mdpa file. This was working in my case, but I want to emphasize it again. Maybe for the future a separate button ("Write MDPA") would make sense in addition to the calculate button.

A last idea of mine is having the possibility to change the name of the submodelpart. Also this would make our lives easier.
Here is a proposal for the structure of the "Add Submodelpart" branch:
```
group: domain
      name submodelpart: GENERIC_domain (=> default)
      Write elements: True
            Name: Element2D2N (=> default)
      Write Conditions: False
```

These comments are for changes in the future, I will approve this pull request now

RiccardoRossi:
Hi,
     i put @jcotela  in the loop  and I give my 2 cents:

the nodes SHOULD be written **EXACTLY ONCE**. writing it more than once is definitely not a good idea. Right now kratos detects it and prevents a crash, however it goes in a slower reading path and it does a distance check to ensure the node really exists.

Why don't you simply write **all the nodes** once at the beginning? _Personally_ I would gladly pay a performance price for correctness

same things for elements and conditions. right now it is directly an error if you write twice the same element. We could remove this limitation if a decision is taken in this sense, however from my point of view this is **not** a good way forward (see #80 , @KratosMultiphysics/technical-committee let's decide if we allow it)

other than that when you write an element and condition, the name should depend as a very minimum on the number of nodes and on the dimensionality

for example
Element3D4N

means a 3D element with 4 nodes (a tet)

if you want a shell like thing, than it would be
**Surface**Element3D4N (4 nodes in 3D ... a quad)

this information is important since it is used internally in the  construction phase to pick the geometry to be employed. A mismatch in some of the quantities will cause the Kratos to rise an error at the moment of reading

philbucher:
@RiccardoRossi I am confused now.
From what I saw @jginternational writes everything once at the beginning and assigns it then in n Submodelparts, or am I wrong?
Having the same entities in several modelparts is not a problem or is it?

> other than that when you write an element and condition, the name should depend as a very minimum on the number of nodes and on the dimensionality
> 
> for example
> Element3D4N
> 
> means a 3D element with 4 nodes (a tet)
> 
> if you want a shell like thing, than it would be
> SurfaceElement3D4N (4 nodes in 3D ... a quad)
> 
> this information is important since it is used internally in the construction phase to pick the geometry to be employed. A mismatch in some of the quantities will cause the Kratos to rise an error at the moment of reading

@RiccardoRossi what do you mean with this?
It it done correctly now, I just think it would be a good idea to be able to modify the elements/conditions directly from GiD without having to modify the mdpa manually
Of course if I as developer put a wrong name Kratos will throw an error, but what is the problem with that?

jginternational:
Ok, I understand your request @philbucher .
you can always call the function manually from GiD:
Files -> Export -> Calculation file
(Don't worry, GiD expects a.dat file, just give it a name and the problemtype generates the mdpa and project parameters file)

jginternational:
Answering to @RiccardoRossi , nodal coordinates are written only once, at the beginning. What we write for each submodelpart are the nodes of the submodelpart, but only the id, not the coordinates

RiccardoRossi:
hi @philbucher 

**Creating** it more than once is problematic. We still did not decide if we want to allow this (my personal vote is against, however it is technically doable).

once you created an element/condition/node correctly you can then **assign** it to as many model parts as you like. 

to make myself clearer:

a node is created when you give 
   
    Begin Nodes
        1 0.0 0.0 0.0
    End Nodes

similarly a condition/element is created once you provide its connectivity. For example if you write in the mdpa:

    Begin Elements Element2D3N
        1 0  1 2 3
        2 0  3 4 5
    End Elements

    Begin Elements Element2D3N
        2 0  3 4 5
    End Elements

you are attempting to create twice the element with id 2.

The point is that while you may accept the one above (twice EXACTLY THE SAME element) you definitely (again my opinion) do not want to admit 

    Begin Elements Element2D3N
        2 0  3 4 5
    End Elements

    Begin Elements OtherElementType2D3N
        2 0  3 4 5
    End Elements

where "Element2D3N" is overwritten y "OtherElementType2D3N"

philbucher:
@RiccardoRossi 
I think we misunderstand each other. **Of course I do not want to duplicate neither nodes, nor elements or conditions!**
I would like to do the following:
Imagine you create a model with GiD, this is the output:
```
Begin Elements Element2D3N
    2 0  3 4 5
End Elements
```
Now lets say I am developing the "MyAwesomeElement2D3N" - element
Then each time I update the model I manually have to exchange in the mdpa file "Element2D3N" to "MyAwesomeElement2D3N"

So it would be nice to be able to modify the names of elements/conditions that GiD writes in the mdpa file. (default in the example is "Element2D3N", which I would like to EXCHANGE with "MyAwesomeElement2D3N")

So in the end the mdpa file would look like this:
```
Begin Elements MyAwesomeElement2D3N
    2 0  3 4 5
End Elements
```

I hope I made myself clear now
(see my second post)

RiccardoRossi:
Hi @philbucher,

i think i got you right. However there are serious issues with this and the SolidMechanicsApplication, as the same "geometry" may be employed multiple times as part of different elements.

We are discussing internally how to try to make this more clear and robust

philbucher:
Hi @RiccardoRossi ,
I think now I begin to understand.
So you are saying that you create multiple elements on the same (real, not Kratos!) geometry, right?
And that is why you wouldn't want this possibility of changing the names from GiD globally?

RiccardoRossi:
Hi Philip,

yes, the issue is that you may want to assign within GiD different element types to the same geometry.
an example of this is the case in which you want to have different loading conditions on the same geometry. This is implemented inside kratos by superimposing conditions that share the same geometry.

now the problem is that the mechanism of doing this within GiD is delicate, for this reason what we are discussing is the possibility of bringing it on the kratos side of things.

the idea we are working on right now would be to call, from inside kratos, a function like

        AssignElement( "element_type", geometry_list, ASSIGN_NEW_ID)

or

        AssignElement( "element_type", geometry_list, ASSIGN_NEW_ID)

this way the GiD would have only to write to mdpa the element connectivity once, organized in submodelparts. It would be then the kratos task to assign a given element type to a given modelpart, deciding explicitly if new conditions need to be created or what is to be done. 
Anyway...this is only a working idea for now, however it would solve a lot of the problems we have currently with the GiD interface. At the same time it would also minimize the knowledge needed "on the gid side" to write something for kratos.

having said this, there is a process to do what you need

KratosMultiphysics.ReplaceElementsAndConditionsProcess(self.main_model_part, self.settings["element_replace_settings"]).Execute()

which tells which elements are to be replaced by what

philbucher:
Hi @RiccardoRossi,
Now it is clear, thank you for the explanation!

jginternational:
@RiccardoRossi I've added this feature to Fluid and Structural in development mode. Merging to master today if you agree

PR #108, opened 2017-07-16
Updating Python scripts

rubenzorrilla:
Python main scripts have been updated to make them as similar as possible between applications. A unique python script is used for both serial and parallel execution.

### DIFF
--- a/kratos.gid/apps/EmbeddedFluid/python/KratosFluid.py
+++ b/kratos.gid/apps/EmbeddedFluid/python/KratosFluid.py
@@ -80,12 +80,11 @@ for i in range(ProjectParameters["gravity"].size()):
     FluidModel.AddModelPart(main_model_part.GetSubModelPart(gravity_part_name))
 
 ## Print model_part and properties
-if(echo_level > 1):
-    if (parallel_type == "OpenMP") or (mpi.rank == 0):
-        print("")
-        print(main_model_part)
-        for properties in main_model_part.Properties:
-            print(properties)
+if (echo_level > 1) and ((parallel_type == "OpenMP") or (mpi.rank == 0)):
+    print("")
+    print(main_model_part)
+    for properties in main_model_part.Properties:
+        print(properties)
 
 ## Processes construction
 import process_factory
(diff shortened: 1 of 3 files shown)


kratos.gid/apps/Structural/python/KratosStructural.py
```diff
@@ -1,110 +1,88 @@
...
-        used_time = time_fp - time_ip
-        print("::[KSM Simulation]:: [ %.2f" % round(used_time,2),"s", process," ] ")
-
-#### TIME MONITORING END ####
-
-#### PARSING THE PARAMETERS ####
-
-# Import define_output
+from KratosMultiphysics import *
+from KratosMultiphysics.ExternalSolversApplication  import *
```
loumalouomega: Why import whit *?

rubenzorrilla: A time ago we decided to do the import with * in the main scripts.

RiccardoRossi: i think that the argument was "simplicity". However in the solvers it should be used without the *

kratos.gid/apps/EmbeddedFluid/python/KratosFluid.py
```diff
@@ -76,11 +80,12 @@
...
-if(verbosity > 1):
-    print("")
-    print(main_model_part)
-    for properties in main_model_part.Properties:
-        print(properties)
+if(echo_level > 1):
+    if (parallel_type == "OpenMP") or (mpi.rank == 0):
+        print("")
+        print(main_model_part)
+        for properties in main_model_part.Properties:
```
RiccardoRossi: why do we want to print this? is it for debug purposes?

rubenzorrilla: Probably we can remove this one, or maybe print it when echo_level is larger than one (as is done with other print statements).

kratos.gid/apps/Structural/python/KratosStructural.py
loumalouomega: OK

jginternational:
Will this work with the [Release](https://github.com/KratosMultiphysics/Kratos/tree/Release-5.1) branch?

rubenzorrilla:
Indeed, this must be in the release branch.

Issue #76, opened 2017-04-10
Integration of DEM application 


jginternational:
For the new release, a decision must be taken about DEM.

- OPTION A: Create a new app 
Create a single app for the DEM application
- OPTION B: Pack DEM as a single problemtype
Deploy this new problemtype with GiD internet retrieve system
Add the logo in this interface and launch the new problemtype

In both options, the compiled Kratos should be the same

@KratosMultiphysics/dem 
@KratosMultiphysics/technical-committee

maceligueta:
In technical terms, I prefer Option A. However, the team @KratosMultiphysics/dem is fully involved in several projects right now and we lack resources for writing a new GUI. In terms of workforce resources I would choose Option B, as the GUI we are using now is fully operational.

jginternational:
I rather prefer Option A. I can help you providing a template of a basic application, so you just need to add things to the xmls (elements, conditions...), and giving support for the tree login and the files writting, but unfortunately all the implementations related to the project parameters written in python have been deprecated, they are not longer supported by the common code, so an effort must be done there (reading json instead of writing python).

Option B can be faster to achieve, but keep in mind that will remain as a separated problemtype.

pooyan-dadvand:
@jginternational How is going with DEM interface?

jginternational:
@maceligueta If we want to discard the "old" DEM integration and create a new app in the current problemtype, we'll need to define some things, so I'll create a new issue for that

jginternational:
Alle fine, we have defined a draft for the DEM app.
@KratosMultiphysics/dem 

Dem app will always be in 3D, Dynamic, and explicit, and will use the Symplectic Euler integration scheme.
Only available in OpenMP, with no linear solver.

In the parts section: there is only one element type. To start with, we will force to mesh this volumes with **spheres**. There will be a materials list and a **Constitutive law list**

In the **boundary conditions** section:
- Inlets: surfaces, lines or points : uses intervals
- Velocity and angular velocity:  uses intervals

In the **conditions** section:
- Walls: surfaces (triangles)

**Results**:
- Elemental (Particles)
- Walls

For the **write** procs, there will be 4 mdpa files and a projectparameters.py
- mdpa1 (parts)
    - Properties
    - Nodes (only parts -> inefficient)
    - Elements (Spheres, will need to add this format to the accepted - eid pid nid rad)
    - Submodelparts
        - Elements
        - Velocity and angular velocity

- mdpa2 (walls)
[...]

PR #122, opened 2017-07-25
Add dem application link

jginternational:
Solves #76 by calling the old DEM Problemtype. 

Waiting for the @KratosMultiphysics/technical-committee to approve


pooyan-dadvand:
As I understood this for using the old interface as it is the only functional one at this moment. 👍

Issue #159, opened 2017-10-17
Updating StructuralMechanics Solvers


philbucher:
Hi @jginternational 
Here I list the changes that should be made to the problemtype for PR [#850](https://github.com/KratosMultiphysics/Kratos/pull/850)
All the changes are under `Model Information`. These are the settings that should be there from now on, for static and dynamic analysis:

Solver Type : Static (formerly: Solution Type)
Solution Strategy: No longer needed
Scheme: Also not needed
Analysis type: leave as is

Solver Type : Dynamic (formerly: Solution Type)
Time Integration Method : Implicit (=> other options will be added in the future)
Scheme: Leave as is
Analysis Type: Same as for Static, i.e. Linear or Non-Linear

These settings also have to be written to the ProjectParameters, under `solver_settings`:
```
"solver_type"     : "static" / "dynamic"
"analysis_type"  : "linear" / "non_linear"
in case of Dynamic:
"time_integration_method" : "implicit"
"scheme_type" : "newmark" / "bossak"
```

Another minor change was done with the naming of the convergence criteria; they are all lower case with underscores, according to the [style Guide](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide)
Thanks a lot!

philbucher:
@jginternational the changes are merged now, please feel free to apply the modifications to the ProblemType. Or show me where to do it, then I can try to do it myself.

philbucher:
Also we removed the old beam `SmallDisplacement...`, it is replaced by `CrLinearBeamElement3D2N` and `CrBeamElement3D2N`
Furthermore we added two more shell elements: a Quad `ShellThinElementCorotational3D4N` and a Triangle `ShellThickElementCorotational3D3N`

philbucher:
And there are now also trusses: `TrussLinearElement3D2N` and `TrussElement3D2N`

rubenzorrilla:
@philbucher I'm tackling the new structural solver inconsistencies.

rubenzorrilla:
@philbucher, @jginternational  and me (together with @loumalouomega ), think that its better to keep the old camel format to avoid breaking the interface as much as possible. Thus, @loumalouomega is doing a PR changing the structure solvers wrapper according to the next convention:

- "solver_type"     : "Static" / "Dynamic"
- "analysis_type"  : "Linear" / "NonLinear"
in case of Dynamic:
- "time_integration_method" : "Implicit"
[...]

PR #163, opened 2017-10-24
Updating structural mechanics solvers#159

rubenzorrilla:
Dear all,

Finally we manage to keep all your changes except the "dynamic"/"static" field, which values are "Dynamic"/"Static" (the main reason is to keep the stable interface version until the released is packaged #159).

@loumalouomega is doing a PR in the StructuralMechanicsApplication to include such "Dynamic"/"Static"  modifications.

Hope this is ok for you.

### DIFF
--- a/kratos.gid/apps/FSI/examples/MokChannelWithFlexibleWall.tcl
+++ b/kratos.gid/apps/FSI/examples/MokChannelWithFlexibleWall.tcl
@@ -365,7 +365,7 @@ proc FSI::examples::TreeAssignationMokChannelFlexibleWall {args} {
     }
 
     # Structure Bossak scheme setting
-    spdAux::SetValueOnTreeItem v "Bossak" STScheme
+    spdAux::SetValueOnTreeItem v "bossak" STScheme
 
     # Structure domain strategy settings
     set str_change_list [list residual_relative_tolerance "1e-8" residual_absolute_tolerance "1e-10" max_iteration "20"]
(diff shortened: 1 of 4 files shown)


philbucher:
Hi Guys,
Thx for your effort!
@rubenzorrilla could you please also add the new elements that I described in issue #159 ?

@jginternational I see your point, but I also think that at some point we should stick to (our own) conventions. It is a developer verion

philbucher:
Before I forget, should we change this after the release then?
Or are there too many dependencies somewhere else?

rubenzorrilla:
@philbucher I can add the shells tomorrow morning. Regarding the beam and the truss elements, I don't know if the interface is already prepared for including them...

philbucher:
@rubenzorrilla thx
there was a beam already, which has been replaced, so I guess the interface is there.

rubenzorrilla:
Ok. Lets ping @jginternational for the beams issue.

jginternational:
As far as I know there is a BeamElement in the interface, but we need to improve it (section type?, rotation angle from the local axis convention...)
@rubenzorrilla we can work on it tomorrow morning, because then I'll be out until next week

philbucher:
@jginternational there is an old beam element, but it was removed from Kratos already.
@KlausBSautter is working on testing the local axis workflow you recently implemented.
However the truss elements you can add for sure, they don't need the local axis feature

PR #181, opened 2017-11-04
update logos and custom results file to Solid

josep-m-carbonell:


josep-m-carbonell:
@jginternational do not merge it yet I want to add more changes in icons, a similar style as pfem for the solidapp. But I have encountered a couple of problems. The icons for group is not changing in the pfem (I remember that they did before)   and I don't know how to set it also for the solidapp. Also the icon for Linear Solver I don't know how to change it.

kratos.gid/kratos_default.spd
```diff
@@ -8,12 +8,12 @@
 	<hiddenfield n="activeapp" active="0" v="" state="disabled"/>
 	<value n="nDim" pn="Dimension" un="nDim" v="undefined" state="disabled" values="undefined,2D,2Da,3D" dict="undefined,Undefined,2D,2D,2Da,2D axisymmetric,3D,3D"/>
 	<appLink n="Solid" public="1" pn="Solid" active="0" appid="Solid" prefix="SL"/>
+	<appLink n="Pfem" public="1" pn="Pfem" active="0" appid="Pfem" prefix="PFEM_"/>
```
jginternational: There is no need for this change

josep-m-carbonell: I put the two applications that I manage together. If you prefer you can move them whenever you want.

josep-m-carbonell: I have restored the position of the Pfem app

PR #183, opened 2017-11-08
changing the view default of icons Pfem/Solid apps

josep-m-carbonell:


josep-m-carbonell:
I have also added some corrections

PR #205, opened 2017-11-28
Setting PYTHONHOME

roigcarlo:
This will fix the "no platform libraries detected for xxx" error


roigcarlo:
P.s. In linux. In windows works ok

PR #239, opened 2018-01-17
StructralMechanics Removed unnecessary Parameters

philbucher:
This PR removes the `restart_options` and the `constraints_data` from the Projectparameters

The restart options will be in the solver soon (I am working on this atm), the constraints data is not needed bcs this is done via processes

This is the first step of cleaning the Problemtype for release 5.3


rubenzorrilla:
:+1:

PR #242, opened 2018-01-26
[StructuralMechanics] Adding new Elements

philbucher:
This PR adds the prestress membrane elements to the interface
Also the shells can now be used both for tri and quad meshes, one does not have to select the geometry type (now it works the same as for the solids)

Step by step doing the stuff from the [TODO-List](https://github.com/KratosMultiphysics/GiDInterface/projects/9#column-1666919)


philbucher:
Ping @loumalouomega

PR #264, opened 2018-02-21
[StructuralMechanics] Updating elements after latest commits

philbucher:
This PR updates the elements in StructuralMechanics after https://github.com/KratosMultiphysics/Kratos/pull/1452

Also I reordered the elements based on their dimensionality, I think this way it makes more sense

@KlausBSautter can you please test it, then we can merge it before the release


philbucher:
@jginternational we have a problem when selecting the ConstitutiveLaw for the Beam: We duplicated it to have a 2D and 3D version but for some reason the filtering is not working, also in 3D we always get the one for 2D
Help is appreciated :)

pooyan-dadvand:
I understand that it is for release 5.3. Isn't it?

philbucher:
not absolutely necessary but would be a nice addition / enhancement / cleanup

also it adds some postprocessing features for the beam

RiccardoRossi:
@pooyan-dadvand this completes output on gauss points for beam diagrams, so i give my +1 to merging it prior to release

philbucher:
There is one small problem that I described above for which I need help from @jginternational , but other than that it is ready to be merged

kratos.gid/apps/Structural/xml/ConstitutiveLaws.xml
```diff
@@ -78,23 +78,37 @@
...
-            <parameter n="KratosMultiphysics.StructuralMechanicsApplication.TORSIONAL_INERTIA" pn="Torsional inertia" type="double" v="1.0" unit_magnitude="L^4" units="m^4" />
-            <parameter n="KratosMultiphysics.StructuralMechanicsApplication.I22" pn="Inertia 22" type="double" v="1.0" unit_magnitude="L^4" units="m^4" />
-            <parameter n="KratosMultiphysics.StructuralMechanicsApplication.I33" pn="Inertia 33" type="double" v="1.0" unit_magnitude="L^4" units="m^4" />
+            <parameter n="CROSS_AREA" pn="Cross area" v="1.0" unit_magnitude="Area" units="m^2" />
+            <parameter n="I33" pn="Inertia 33" type="double" v="1.0" unit_magnitude="L^4" units="m^4" />
+        </inputs>
+        <outputs></outputs>
+    </CLaw>
+
+    <CLaw n="BeamConstitutiveLaw" pn="Linear elastic beam" ProductionReady="ProductionReady" Type="Beam3D" behaviour="Elastic"
```
jginternational: @philbucher The 'n' field is the Primary Key for the constitutive law database, so 2 CLaw can not have the same n

In this case, I see 2 CLaw with  n="BeamConstitutiveLaw"

philbucher: Yes because the name is the same for 2D and 3D but the variables to be written are different
I checked the elements, which filter the ConstLaw by `Type`, wich I changed to `Beam2D` and `Beam3D` respectively

jginternational: Well, they must have different names if they're different. The only use case for using twice the same name is to replace the previous one.

jginternational: Or, we could think about a tricky solution, if they must be named with the same way

philbucher: I don't think it's worth it, I will revert the change

philbucher: I just saw that `LinearElasticPlaneStress2DLaw` exists three times with the same name, how does this work?
Isn't this the same I am trying to do?

jginternational: Nice discovery, It is not working, properly, and it's almost impossible to debug.

I was working on a local branch to add a "Kratos_Name" tag to the constitutive laws, so the 'n' becomes into an identifier only relevant to the GUI, and the printed name can be repeated.

Let me try to merge it before this pull request and let's check if it works

jginternational:
I think it's fixed.

philbucher:
yep it works, thx @jginternational

PR #266, opened 2018-02-26
Properties id consecutive for DEM files

### LINKED ISSUE #254: Dem properties must have unique id
DEM application writes 4 mdpa. Each one is starting it's properties from 0.

Those ids should be unique globally, so if 'parts mdpa' properties are 0, 1, 2 and 3, Inlet properties should be 4 and 5.

Relates #76

jginternational:
Fixes #254

### DIFF
--- a/kratos.gid/apps/DEM/write/write.tcl
+++ b/kratos.gid/apps/DEM/write/write.tcl
@@ -21,9 +21,8 @@ proc DEM::write::Init { } {
     variable inletProperties
     set inletProperties [dict create]
     
-    # Initialized as -1, so next assigned gets id 0
     variable last_property_id
-    set last_property_id -1
+    set last_property_id 0
 }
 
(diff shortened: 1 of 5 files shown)


jginternational:
What do you mean with 

> the properties are not initialised.

?

merceln:
The properties ID, sorry. To be more precise, when I run the case the second time, the Parts properties IDs are initialised to 1, but the Inlet properties Id and walls properties ID continue the numbering of the previously launched case. 
id numbering

jginternational:
@merceln I think it's fixed

merceln:
@jginternational the first time I run a case it works fine now, but when I launch it the second time without closing the problemtype, the counter of last_property_id is not initialised to 0 (DEM::write::Init proc is not invoked), so the properties id continues the numbering of the previous case launched.

PR #278, opened 2018-03-07
update solid and pfem apps

josep-m-carbonell:


josep-m-carbonell:
thank you @jginternational

PR #280, opened 2018-03-09
This line is giving me problems when it checks the type of element

loumalouomega:


loumalouomega:
OK, I solve the problem, apparently there is a problem with the names of groups in TCL, I redid the groups and it works as always, sorry for the noise, but maybe is a real bug

PR #288, opened 2018-03-10
Enhacement change all mesh to submodelpat

### LINKED ISSUE #187: Remove MDPA references to Mesh
you shouldn't write mesh_id=0 in the process definition.

It does not harm, but it is not needed

jginternational:
Fixes #187


jginternational:
@RiccardoRossi Al references to Mesh in MDPA or json have been deleted

PR #296, opened 2018-03-14
Addition of Bbar element and J2 constitutive model to the interface

rubenzorrilla:
@KratosMultiphysics/structural-mechanics team, I've added the Bbar element together with the J2 constitutive model. However, I still have some doubts so I'm opening the PR to start the discussion.

- the Bbar elements can only work with the J2 cons. law or can we use other cons. laws?
- default values for the REFERENCE_HARDENING_MODULUS, REFERENCE_HARDENING_MODULUS, INFINITY_HARDENING_MODULUS and HARDENING_EXPONENT. Recall that in the interface we have as default materials steel and aluminium, so it would be nice if you can provide these magnitudes for these materials.
- are you interesting in any output variable different to the common ones?

### DIFF
--- a/kratos.gid/apps/Structural/xml/ConstitutiveLaws.xml
+++ b/kratos.gid/apps/Structural/xml/ConstitutiveLaws.xml
@@ -53,10 +53,10 @@
             <parameter n="DENSITY" pn="Density" unit_magnitude="Density" units="kg/m^3" v="7850"/>
             <parameter n="YOUNG_MODULUS" pn="Young Modulus" unit_magnitude="P" units="Pa" v="206.9e9" />
             <parameter n="POISSON_RATIO" pn="Poisson Ratio" v="0.29"/>
-            <parameter n="YIELD_STRESS" pn="Yield stress" unit_magnitude="P" units="Pa" v="278.0e9"/>
-            <parameter n="REFERENCE_HARDENING_MODULUS" pn="Kinematic hardening modulus" unit_magnitude="P" units="Pa" v="1.0"/>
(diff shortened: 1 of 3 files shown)


marandra:
Hi @rubenzorrilla 

> the Bbar elements can only work with the J2 cons. law or can we use other cons. laws?
The 3D element could be used with any law, and the results will be very close, but not equal, to the results of the regular small strain element (of course the definition of "close" depends on the application). The use case for the B-bar element is to avoid volumetric locking that may appear in some situations, especially when using plasticity. The use case for the 2D element is the same, with the addition that it considers the zz component of the strain tensor, so 4 components, as required by the plasticity 2D CL. It makes it incompatible with the other 2D CLs, as they consider only 3 components.
Also, note that the B-bar elements are defined for hexahedral (8N3D) and quadrilateral (4N2D) geometries only.

rubenzorrilla:
Summarizing,

- 3D Bbar element: any CL.
- 2D Bbar element: only J2 plan strain law.

Regarding the geometries, yes. It can only be meshed with quads and hexas.

Any suggestion for the J2 CL defaults? IMO it is the only thing missing before starting to test it.

marandra:
Hi Ruben
I think we can use the parameter values used in the test case as the default ones in the GUI for now. Later we can update them to something consistent with the other models.
```
YOUNG_MODULUS, 21000
POISSON_RATIO, 0.3
YIELD_STRESS, 5.5
REFERENCE_HARDENING_MODULUS, 1.0
ISOTROPIC_HARDENING_MODULUS, 0.12924
INFINITY_HARDENING_MODULUS, 0.0
HARDENING_EXPONENT, 1.0
```

rubenzorrilla:
Done!

PR #305, opened 2018-03-21
Feature add eigensolveranalysis

### LINKED ISSUE #243: Structural - Add Eigensolveranalysis
As said in KratosMultiphysics/Kratos#1458

This issue will contain the GiD needs to achieve the required implementation

jginternational:
Fixes #243


philbucher:
looks good to me ecxept that it loads the wrong output file when switching to PostProcessing => described in my second post in #243 
In the same line you don't need to write `output_configuration` to ProjectParameters bcs the EigenValues cannot be displayed with the regular output

also `analysis_type` is still displayed in the ProjectParameters

Last but not least could you please hide the SchemeType "Stability" for now, this is for the future

thx @jginternational

philbucher:
Now the analysis works, but it does not load the results when I switch from pre to post. I think it loads the .msh file where it should load the .bin file

jginternational:
Do it works if you load the postprocess file manually? @philbucher

philbucher:
yes

jginternational:
The results file should be named "model_name.post.res" or "model_name.post.bin" and it will be opened automatically

PR #343, opened 2018-05-11
Updating to match changes in Fluid application.

jcotela:
This updates the interface to match PR 1904 in Kratos.


jcotela:
Well, if you look at it, it is already a launcer script (with a trick to auto-flush every 10 seconds). I'd love to never have to change the main script for the fluid again... Unfortunately I was already hoping that this would be case the last time I changed it, but now it has been decided that the main script must accept the Model as argument (which is the change I did here). I hope I can keep it constant for a while now, @KratosMultiphysics/technical-committee willing...

PR #350, opened 2018-05-17
[Structural] removing old membrane

philbucher:
see [PR in Kratos](https://github.com/KratosMultiphysics/Kratos/pull/2113)


loumalouomega:
I trust @jginternational

philbucher:
Nope, we’re replaced by the prestressmembrane (which is already in the interface)

Issue #335, opened 2018-04-27
Check proc for writing Solvers in the tree and establishing default values of each param


djvicente:
Hi @jginternational,
I have detected one possible bug related to the selection of the 'Solver' and the default values which appear in the tree. I will try to explain it. In the DamInterface I have solved this bug, but I think it could be interesting to comment it. 

I needed to change the default value of one parameter of a specific Solver (in my case the 'BICGSTABSolver'). In my app, I read the 'Common/xml/Solvers.xml' and introduce the Solvers in the tree with the proc 'spdAux::injectSolvers' through a 'dynamicnode'.
I have found that, although the default value of a specific parameter in a specific solver is changed (in my case 'preconditioner_type --> v="ILU0Preconditioner", if the same parameter is common in another solver written after the one I have changed, the default value is set into the value of the last solver.

I think, but I am not sure, that the origin of the problem is the way it is imported the Solvers to each through the spdAux::injectSolvers proc.

To check this bug I have made some tests. For example, changing all the default values of a common parameter, 'max_iteration'. I put a different value for each solver and it always take the default value writen in the last solver 'AmgclMPISolver'. If I change the order of the solvers, it still takes the last one.

In my case, I have solved my problem copying the 'Solvers.xml' file, and writing my default solver with the desired params and values at the end of the Solvers List. This action solves my problem, but maybe it can be fixed in a global way. 

I hope that it will help you discover this bug and if I have not explained well, do not hesitate asking me

jginternational:
This is a well known bug. There was an discussion about it a couple years ago. There was a problem with the default values here with the parameters 'shared' between different solvers. 

Imagine you have Solver A and Solver B, and both of them havethe param Parameter
For Solver A, the default value of the param is 1
For Solver B, the default value of the param is 2

The first time a user opens the tree, Solver A is selected, and Parameter value is 1
The user changes this value to 5
Then changes the solver to Solver B.
What should Parameter value be? 2 or 5?

It's not easy. For the next Big change, expected to be published after summer, this will be solved.
For this case, Parameter value will be 2, but if you change again to Solver A, it will remember the 5

djvicente:
Perfectly explained @jginternational. Then, as you were aware of this aspect, I close this issue. 
Thank you very much for your help

PR #357, opened 2018-05-24
Add values and proc to 'file' type parameter

djvicente:
'spdAux::GetParameterValueString' has been modified in order that, if developers use type 'file' for certain processes, external files can be easily linked. The attributes included are the same as such introduced for the 'tablefile' type. I have tested it in the DamApplication and it works perfectly.

### DIFF
--- a/kratos.gid/scripts/spdAuxiliar.tcl
+++ b/kratos.gid/scripts/spdAuxiliar.tcl
@@ -5,15 +5,15 @@
 
 namespace eval spdAux {
     # Namespace variables declaration
-    
+
     variable uniqueNames
     variable initwind
-    
+
     variable currentexternalfile
     variable refreshTreeTurn
-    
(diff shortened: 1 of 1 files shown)


djvicente:
Hi @jginternational,
I have just started a pull request for a change in the GidInterface. I have introduced this change in the DamApp and it works perfectly. As you can see, is a minor change, that can help other developers who work with external files. 
By the way; did you have the opportunity to check this Issue? https://github.com/KratosMultiphysics/GiDInterface/issues/335
If I can help you do not hesitate to contact me.

djvicente:
@jginternational, it does work with the '-' option! I thought that the '-' command meant that such option did nothing. I should have tried it before modifying it. Therefore, I think it's not worth changing it.
Sorry for the mistake... and thank you for checking it.

PR #361, opened 2018-05-28
Optional parameters protection #360

### LINKED ISSUE #360: Error due to changes in Writing.tcl
Hi @jginternational 
One of the last changes introduced in the Writing.tcl leads to error (Find attached a screenshot of the error window message)
I think it is related to the changes introduced in lines 1086-1090. To fix it, I have commented the line 'dict unset process_attributes process_name' and it works. I hope this may help you. 

error_writing

jginternational:
Fixes #360

### DIFF
--- a/kratos.gid/scripts/Writing.tcl
+++ b/kratos.gid/scripts/Writing.tcl
@@ -1086,8 +1086,8 @@ proc ::write::getConditionsParametersDict {un {condition_type "Condition"}} {
             dict set process_attributes process_name [dict get $process_attributes n]
             dict unset process_attributes n
             dict unset process_attributes pn
-            dict unset process_attributes help
-            dict unset process_attributes process_name
+            if {[dict exists $process_attributes help]} {dict unset process_attributes help}
+            if {[dict exists $process_attributes process_name]} {dict unset process_attributes process_name}
             
             set processDict [dict merge $processDict $process_attributes]
             if {[$condition hasAttribute VariableName]} {


djvicente:
Hi Javi,
Thank you for your quick response, but I think it does not solve my problem. 
With your last change, the attribute 'process_name' is firstly set (LINE 1086) and then unset (LINE 1090). I think that the new command "if {[dict exists $process_attributes process_name]} (...)" does not make sense, given that the attribute process_name is always previously set, and hence it always exists. Maybe, with the 'help' attribute this command makes sense, as it can be defined or not in the Processes.xml file.

The question is: it is necessary to delete/unset the attribute 'process_name'? In the DamApplication, certain processes depend on this attribute. Maybe, I would be able to change those processes, but I'd rather keep them.

What do you think?

jginternational:
Hi @djvicente 

process_name and help are fields that, as n or pn, belong to the GUI. They should 'not cross the line' to the python/c++.  

Your script should work using the fields 'python_module' and 'kratos_module'

djvicente:
Hi again @jginternational,
I agree with you that those fields belong to the GUI and shouldn't be transferred to Kratos. When I said that certain processes depended on the process_name field I was referring to certain 'procs' of the WriteProjectParameters.tcl of the DamApplication.
Anyway, if you think that it is necessary to delete them, I will change my code considering the 'python_module' attribute as input variable of my procs instead of 'process_name'.
Thank you Javi,

PR #384, opened 2018-06-01
[MPM] Removing useless lines from MainKratos.py

bodhinandach:


jginternational:
Hi @bodhinandach 
Will this work with the [Release 6.0.0 binaries](https://github.com/KratosMultiphysics/Kratos/releases/tag/v6.0-Beta2)?

bodhinandach:
@jginternational it will work since we just remove lines to call some (deprecated) python functions.

PR #386, opened 2018-06-01
Corrections for the Release Solid App

josep-m-carbonell:
Solver truncates the precision of the examples


josep-m-carbonell:
I want to change the default values... can I do it without duplication the Solvers.xml?

jginternational:
You can use the CustomTree section of the GetFromXML script

josep-m-carbonell:
Ok, I added also a method to forget a solver.

Issue #320, opened 2018-03-30
To do in the PFEM-fluid interface


AFranci:
1) Constitutive Laws still makes reference to PFEM interface, not PFEM-fluid.
2) When FSI is selected as Physics for the first time, the elements that then appear are not those of the PFEM fluid. In case of selecting before a Fluid physics and then again the FSI one, the problem disappears and one can choose normally the PFEM-fluid elements. 
3) When a fluid/solid body type is selected, then it would be great to see just fluid/solid elements. For example, if a fluid  body type has been chosen, only Newtonian and non-Newtonian elements should appear in the element list. On the contrary, if the body is Solid, the only options should be Hypoelastic.
4) When the reults are opened in the post process, it appears the warning: libpng warning: iCCP: knwown incorrect sRGG profile (You already told me that this is not your issue but I put it here anyway to not forget it...)

AFranci:
point (4) solved by Miguel!

PR #387, opened 2018-06-01
Constitutive taken from the PFEM fluid for fluid elements

AFranci:


AFranci:
Please take a look to the points 2 and 3 of issue #320

PR #401, opened 2018-06-19
Feature transform from older version

### LINKED ISSUE #302: Transform between versions is needed
Just to remember it here as a critical thing to implement

jginternational:
Fixes #302

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ The interface of Kratos with [GiD](http://www.gidhome.com)
 * kratos top menu / Developer mode (recommended)
 
 ## Warnings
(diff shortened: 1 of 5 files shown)


jginternational:

Feature transform will only work properly in 

- GiD developer versions > 14.1.1d 
- GiD Official versions > 14.0.1

If I increase the "Minimum GiD version" tag in kratos, users of GiD 13 will no longer work

At this moment, I've placed a warning for those GiD versions older than the minimum, when trying to do a transform.

Should I increase to 14 the "Minimum GiD version"?

Issue #430, opened 2018-07-17
Setting addition instead of assignation in the Boundary Conditions


josep-m-carbonell:
In some cases the condition must add the value to the current value of the variable. 
The question is how to set in the tree that the BC value must be added instead of overwritten.

jginternational:
If you need the user to decide which values must be added and which must overwrite the existent, you can add a boolean parameter (or a combo if there are more options) to the process and the user will choose.

If you don't want the user to choose it, then is the python process the one that should implement that logic .

josep-m-carbonell:
The user must select this option. Add or Overwrite, which other options do you think about?

jginternational:
I was thinking in some kind of priority scale, but maybe it's too much.

You can try and add yourself the parameter in the process

josep-m-carbonell:
It is a good idea, well the priority is the order in which they are defined, the best way would be to be able to interchange the order of the process in the menu's tree.
Right now, the combo can have different options for the compound assign: Assign, Addition, Subtraction, Multiplication, Division. The default is Assign.

jginternational:
This has been implemented in the problemtype but can not be used in master branch of kratos yet

it fails:

```
Traceback (most recent call last):
  File "MainKratos.py", line 7, in <module>
    MainSolid.Solution().Run()
  File "E:\PROYECTOS\KratosGiD\kratos.gid\exec\kratos\applications\SolidMechanicsApplication\python_scripts\MainSolid.py", line 52, in Run
    self.Initialize()
  File "E:\PROYECTOS\KratosGiD\kratos.gid\exec\kratos\applications\SolidMechanicsApplication\python_scripts\MainSolid.py", line 72, in Initialize
    self.processes = self._get_processes()
  File "E:\PROYECTOS\KratosGiD\kratos.gid\exec\kratos\applications\SolidMechanicsApplication\python_scripts\MainSolid.py", line 353, in _get_processes
    return (process_handler.ProcessHandler(domain_model, processes_parameters))
  File "E:\PROYECTOS\KratosGiD\kratos.gid\exec\kratos\applications\SolidMechanicsApplication\python_scripts\process_handler.py", line 35, in __init__
    self.Sort()
  File "E:\PROYECTOS\KratosGiD\kratos.gid\exec\kratos\applications\SolidMechanicsApplication\python_scripts\process_handler.py", line 110, in Sort
    self.list_of_processes  = self.ConstructList( self.ConstraintsList() )
[...]

PR #433, opened 2018-07-25
adding compound assignment combo in nodal conditions

josep-m-carbonell:
Adding compound assignment option for nodal conditions #430


jginternational:
Have you tested it?

I was thinking in a regular parameter in the process, instead of an xml attribute in the condition, because this way is heavier to mantain, but if it works (and does not break anything) go!

josep-m-carbonell:
I implemented this and it works.  However I want to include it also in the loads. I will make some more tests before doing any merge.

PR #440, opened 2018-08-08
Mdpa import mesh

jginternational:
Utility to import an mdpa into the model.

Imports the nodes and the elements, according to Elements and Conditions, and generates groups according to SubModelParts

- It **does not interact with the Kratos tree** 
- It may **not keep the nodes and element id**, (if previous mesh existing in the model)

It can be found in the Kratos menu tab

image

Just mention @KratosMultiphysics/technical-committee so they can be informed of this new feature


pooyan-dadvand:
Great!

It might be very useful!

PR #453, opened 2018-09-11
[Structural] Use GiDOutputProcess-Factory

philbucher:
fyi @KratosMultiphysics/structural-mechanics 

Using the new factory of the GidOutputProcess to be more consistent with the other processes

see #454


philbucher:
yep fully retrocompatible, same as for the processes

philbucher:
more explanation in #454

PR #469, opened 2018-10-04
Removing unnecessary imports

maceligueta:


jginternational:
Hehe... I've been fighting for a week to compile that in windows... Now that I've reached that goal, they disappear

PR #470, opened 2018-10-08
Feature conv diff write json output process

jginternational:
New Output process implemented in convection diffusion app

@RiccardoRossi @philbucher could someone generate an example for this app?

Everytime I try to run it, the following message appears for each time step

`ATTENTION! setting the RHS to zero!`


philbucher:
@jginternational I am not working with this app

Maybe @rubenzorrilla knows how it works

PR #480, opened 2018-10-30
Fsi new write json

jginternational:
I think it's done.

I'm checking it on your branch on linux

@rubenzorrilla


jginternational:
@rubenzorrilla check the Time Integration Scheme parameters like dynamic_tau, oss_switch
Something is happening there.

PR #485, opened 2018-11-09
[Solid] fix symbol draw in 2D 

josep-m-carbonell:
In 2D direction has only two values and the expected direction vector requires three.
 
In the laptop with the 2K screen no symbol is shown.... Do you think there is something wrong related with the resolution scale @escolano?

There is also another thing. When a direction is zero, because it has not been assigned correctly, it fails to do the norm and a big error/warning message appears. It is possible to check it before and add a small warning message saying that direction is wrong and has to be corrected?

Thank you in advance!

### DIFF
--- a/kratos.gid/apps/Solid/symbols/symbols.tcl
+++ b/kratos.gid/apps/Solid/symbols/symbols.tcl
@@ -32,6 +32,9 @@ proc Solid::symbols::draw_selfweight { valuesList } {
         set modulus 1.0
     }
     set direction [split $data(direction) ,]
+    if { [llength $direction] == 2 } {
+        lappend direction 0.0
+    }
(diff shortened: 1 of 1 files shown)


escolano:
I have only written the first version of drawing symbols for the Solid - 3D problemtype (although the 3D symbols are declared to be used also for Solid 2D problemtype). 
As you did, for 2D it is necessary to provide the third coordinate z=0 
And maybe the sames of some fields are different in both cases, then the appropriated name must be set of course (e.g. fields to know the direction and modulus of a force, etc.)
To show these symbols is necessary to have GiD 14.0.2 or 14.1.2d (already not released in the webpage of GiD), because it has been necessary fix some problems of gid.exe to be compatible with some Kratos particularities.

PR #493, opened 2018-11-26
Add bfs cht

jginternational:
@rubenzorrilla Try it please, I can't run it. It's the second example in the toolbar


rubenzorrilla:
The inlet group has no nodes nor conditions. Besides, we need to change its geometry as we discussed before.

Concerning the error raised, it is because of the solver. I would do a PR to fix it ASAP.

On the other hand, we are now writing 3 levels of processes:

1. processes or output_processes
2. fluid or solid processes
3. initial, boundary condition, auxiliary, etc...

This breaks the processes construction in the base analysis_stage.py _CreateProceses() method.

rubenzorrilla:
@jginternational the last changes should circumvent the problems above. I will fix the ones in the Kratos side ASAP.

PR #501, opened 2018-12-17
[Pfem-Solid] Update

josep-m-carbonell:
Update interface related to the Kratos-master update PR


kratos.gid/apps/Pfem/xml/GetFromXML.tcl
```diff
@@ -2,6 +2,7 @@ namespace eval Pfem::xml {
     variable dir
     variable bodyNodalCondition
     variable body_UN
+    variable Elements
```
jginternational: Be careful, I think this variable is unused but I think that this way is dangerous (trying to override core variables)

josep-m-carbonell: I will delete it in another PR.

PR #502, opened 2019-01-04
adding basic vtk-output

philbucher:
using the new vtk-output-process
FYI @KratosMultiphysics/technical-committee @KratosMultiphysics/interface-committee 

so far it only supports nodal-solution-step-results


adityaghantasala:
I have very little knowledge about this, but seems fine. Just one comment : 
Is it possible to set the default output_control_type for the VTKoutput as step ?

philbucher:
ok now I am setting the correct settings by default

I will just merge this and then we see how ppl use it or if there is sth not working

PR #504, opened 2019-01-15
Hotfix cht wrong modelparts write

jginternational:


rubenzorrilla:
Thanks @jginternational !

PR #509, opened 2019-01-28
updating mesh-solver names

philbucher:
To match the current kratos master

I am not sure which names are the correct ones to update though


rubenzorrilla:
I've just realized that the FSI interface is not writing the ProjecParameters.json properly. We will use this branch to also include that fix.

rubenzorrilla:
Works without problems. Thanks for updating.

PR #520, opened 2019-02-04
updating solver-names

philbucher:
I started updating the names of the linear solvers, but it is not complete / tested (I leave that to you since I don't know how)
The names of all linear-solvers can be found in the following files (search for `KRATOS_REGISTER_LINEAR_SOLVER` and `KRATOS_REGISTER_TRILINOS_LINEAR_SOLVER` respectively):
[Core](https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/sources/standard_linear_solver_factory.cpp#L69-L77)
[ExternalSolversApplication](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/ExternalSolversApplication/custom_python/add_linear_solvers_to_python.cpp#L141-L159)
[EigenSolversApplication](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/EigenSolversApplication/custom_python/add_custom_solvers_to_python.cpp#L116-L141)
[TrilinosApplication](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/trilinos_application/custom_factories/trilinos_linear_solver_factory.cpp#L40-L78)
[Preconditioners (only in Core)](https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/sources/standard_preconditioner_factory.cpp#L47-L50)

**Note**: For the solvers defined in the `ExternalSolversApplication` and the `EigenSolversApplication` it is necessary to prepend the name of the Application: e.g. `ExternalSolversApplication.super_lu`
=> this is necessary to import the application if necessary

Let me know if you need help have questions

BTW I am not sure which of those solvers you want to use, that you have to decide yourself

FYI @KratosMultiphysics/technical-committee


rubenzorrilla:
@philbucher now is all in our side. Thanks a lot for the implementation and explanation!

philbucher:
@roigcarlo I think the EigenSolversApplication is now also compiled in the release right?
Then it would be cool to at least include the non-pardiso (aka non-Intel) direct-solvers 

FYI @armingeiser

philbucher:
ping @rubenzorrilla @jginternational

rubenzorrilla:
@philbucher sorry for the delay

kratos.gid/apps/Common/xml/Solvers.xml
```diff
@@ -39,7 +39,7 @@
         </inputs>
     </solver>
 
-    <solver n="AMGCL" pn="AMGCL" SolverType="Standard" Parallelism="OpenMP">
```
rubenzorrilla: Guys, I've just realized that if we rename the `AMGCL` to `amgcl`the GUI takes the Trilinos AMGCL since its name is also `amgcl`. If the serial settings coincide with the MPI ones we can think on having a unique `amgcl`entrancy. Blocking until we ensure that.

rubenzorrilla: As far as I understand the settings are exactly the same since the MPI version is derived from the serial one. Having this in mind, we need to decide which AMGCL features do we want to expose to the user.

philbucher:
this PR was intended to be used as a basis, since I don't know the details of how the implementation works and last time I broke sth

and yes, the names btw serial/MPI have been unified

kratos.gid/apps/Common/xml/Solvers.xml
rubenzorrilla: As far as I understand the settings are exactly the same since the MPI version is derived from the serial one. Having this in mind, we need to decide which AMGCL features do we want to expose to the user.

rubenzorrilla:
No worries, I'm taking care of this.

rubenzorrilla:
@jginternational last commit should temporary fix the AMGCL OpenMP and MPI issue. We need however to find a way to avoid duplicating the solver for these situations in where the same solver works for both OpenMP and MPI. Any suggestion?

philbucher:
guys please merge this when you think it is ready

rubenzorrilla:
@philbucher I would merge it as it is and let @jginternational to find a fix to avoid the AMGCL entry duplication.

philbucher:
sure, please go ahead

PR #555, opened 2019-03-14
[FSI] Turek benchmark settings

### LINKED ISSUE #549: [FSI] Turek example settings
The Turek example in the FSI application settings are not robust enough so that some elements become inverted when large displacement occurs. Besides the solid CLAW is not the one specified in Turek's paper.

rubenzorrilla:
Change Turek benchmark example mesh & settings to ensure convergence.

@jginternational I think that after merging this we can say that the FSI GUI is ready to be released.

closes #549


jginternational:
Gas

Issue #497, opened 2018-12-03
PotentialApplication not running


inigolcanalejo:
Does someone know the solution to the following error?

![gid_error](https://user-images.githubusercontent.com/28628414/49375998-235d9480-f706-11e8-820c-d64dd0c6a4e1.png)

rubenzorrilla:
Hey @inigolcanalejo !

This seems to be related to the conditions writing (we had to modify it in the past and it seems that we forgot to update your application). @jginternational does this make sense to you?

jginternational:
Yes, it's in my TODO list. 

Anyway it should have a running example like the other apps, so this issues will be detected earlier. Check the Fluid toolbar, the Book icon.

inigolcanalejo:
Ok, thanks for the hint and your fast answers. I will prepare the running example.

PR #560, opened 2019-03-20
[PotentialFlowApplication] Working again

rubenzorrilla:
With this PR the GUI of the PotentialFlow application is working again. The Kratos solver and processes have been updated accordingly.

Merging in the main repository is still pending but in any case we can merge this now since it is not working in its current state.

By the way, I deactivated the 3D option since it is neither available in Kratos.


kratos.gid/apps/Fluid/write/writeProjectParameters.tcl
```diff
@@ -3,10 +3,10 @@ proc ::Fluid::write::getParametersDict { } {
     set projectParametersDict [dict create]
 
     # Problem data
-    dict set projectParametersDict problem_data [write::GetDefaultProblemDataDict $Fluid::app_id]
```
rubenzorrilla: @jginternational everything works after this change but please confirm me that it is OK.

kratos.gid/apps/Fluid/write/writeProjectParameters.tcl
```diff
@@ -3,10 +3,10 @@ proc ::Fluid::write::getParametersDict { } {
     set projectParametersDict [dict create]
 
     # Problem data
-    dict set projectParametersDict problem_data [write::GetDefaultProblemDataDict $Fluid::app_id]
+    dict set projectParametersDict problem_data [write::GetDefaultProblemDataDict [apps::getActiveAppId]]
 
     # output configuration
-    dict set projectParametersDict output_processes [write::GetDefaultOutputProcessDict $Fluid::app_id]
```
rubenzorrilla: Same comment.

kratos.gid/apps/Fluid/write/writeProjectParameters.tcl
jginternational: This was set to solve the inheritance problem of calling to common functions from the inherited app.

In this case imagine the FSI

the function apps::getActiveAppId will return FSI but you want to execute the function telling the system to search in the Fluid part of the tree, so this is why $Fluid::app_id must be there

jginternational:
Fixes #497

kratos.gid/apps/Fluid/write/writeProjectParameters.tcl
rubenzorrilla: OK! I'll revert it. Thanks for the explanation.

kratos.gid/apps/Fluid/write/writeProjectParameters.tcl
rubenzorrilla: (to be reverted as well)

rubenzorrilla:
@jginternational required changes done! Just in case, I tried the FSI and works as expected.

jginternational:
As we say in my town, "chachi pistachi"

PR #562, opened 2019-03-25
Fix structural example 3d

jginternational:
@rubenzorrilla 
I understand that the HighRiseBuilding example is only for 2D.

I've hidden it for 3D.
Merge if you agree.
Add it for 3D if you think it should be also in 3D


jginternational:
@rubenzorrilla

PR #580, opened 2019-04-10
Update KratosFluid.py

philbucher:


kratos.gid/apps/Buoyancy/python/KratosFluid.py
```diff
@@ -1,9 +1,7 @@
 from __future__ import print_function, absolute_import, division #makes KratosMultiphysics backward compatible with python 2.6 and 2.7
 
 import KratosMultiphysics
-import KratosMultiphysics.FluidDynamicsApplication
-
-from fluid_dynamics_analysis import FluidDynamicsAnalysis
+from KratosMultiphysics.FluidDynamicsApplication.fluid_dynamics_analysis import FluidDynamicsAnalysis
 
 import sys
 import time
```
rubenzorrilla: I'm wondering if do we need these.

PR #588, opened 2019-04-30
Nurbs modification and 2 wires

soudah:
New Stent Interface

Nurbs 

2 wires


soudah:
@jginternational  @agiuliodori 

We believe that the new "Stent wizard" is ready. 
There is just a small issue with the groups (due to GID tolerance factor). When we create the "nurbs geometry" all the connectors points disappear.

Thanks,
Edu

jginternational:
@soudah @agiuliodori 
Minimum GiD Version 14.1.3d

PR #600, opened 2019-06-05
Fix after mesh generation special for DEM

jginternational:
@farrufat-cimne 
@KratosMultiphysics/dem

The important change is in the ovm field of the Part condition in parts.spd 
was:
`ovm="element"`
and needed:
`ovm="node,element"`

The rest is just stylings and apply default values to things


kratos.gid/apps/DEM/xml/Parts.spd
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<condition n="Parts" pn="Parts" ov="point,line,surface,volume" ovm="element" icon="shells16" help="Select your group" un="DEMParts">
+<condition n="Parts" pn="Parts" ov="point,line,surface,volume" ovm="node,element" icon="shells16" help="Select your group" un="DEMParts">
```
jginternational: This is the important change of this branch
@farrufat-cimne 
@KratosMultiphysics/dem

PR #608, opened 2019-07-01
Mpm/new mpm fem coupling interface

jginternational:
Implementation of MPM - Structure App

Relates to KratosMultiphysics/Kratos#5140

Missing some coupling features


philbucher:
Cool :+1: 
I haven't checked in detail yet but this will be very useful also for other types of coupling in the scope of the CoSimApp, thanks for your efforts

@jginternational just to say that the current coupling is based on an early version of the CoSimulationApp, @bodhinandach and I are currently discussing on how to port the implementation
Hence some changes will become necessary, but not much

bodhinandach:
@jginternational As branch https://github.com/KratosMultiphysics/Kratos/pull/5185 has been merged with master, could we continue on working for the interface?

PR #621, opened 2019-07-18
Parts by 'element family'

jginternational:
I've been working in the assignation of parts of the Structural application
Until now, it was a little bit confusing, as it's explained in the following example. The user didn't know which geometry type was available for each element, and the different elements were mixed in the 'Element' dropdown.

Thes PR improves the UX by grouping the elements in families, so the shells will be together, the beams, the membranes...

### Before
image
All you can see, all the elements are in the same dropdown, and the user must select the geometry type (line, surface, volume) to set the assignation.
Since each element type in this app can only be assigned to a single geometry type, this is a non-sense. Membranes are only available for surfaces, cables to lines....

### New
image
Now, the user 'double-clicks' on Parts > Shells (the element family) and only the shell elements are available in the dropdown. In addition, it can only select surfaces.

I think this is a cleaner design and hope you like it!
Of course, Structural app is being a pioneer. This solution can be easily implemented to any app with the same problem (Fluid dynamic does not need it since the y only have 1 element).

@KratosMultiphysics/structural-mechanics 

### DIFF
--- a/kratos.gid/apps/FSI/examples/HighRiseBuilding.tcl
+++ b/kratos.gid/apps/FSI/examples/HighRiseBuilding.tcl
@@ -164,7 +164,7 @@ proc FSI::examples::TreeAssignationHighRiseBuilding {args} {
     gid_groups_conds::setAttributesF {container[@n='FSI']/container[@n='Structural']/container[@n='StageInfo']/value[@n='SolutionType']} {v Dynamic}
 
     # Structural Parts
-    set structParts {container[@n='FSI']/container[@n='Structural']/condition[@n='Parts']}
+    set structParts {container[@n='FSI']/container[@n='Structural']/container[@n='Parts']/condition[@n='Parts_Solid']}
     set structPartsNode [customlib::AddConditionGroupOnXPath $structParts Structure]
     $structPartsNode setAttribute ov surface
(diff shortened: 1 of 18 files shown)


jginternational:
Missing examples and FSI

philbucher:
Very cool!
Will that also restrict the groups selection?
In the sense that e.g. for shells one can only select groups with surfaces etc?

rubenzorrilla:
> Very cool!
> Will that also restrict the groups selection?
> In the sense that e.g. for shells one can only select groups with surfaces etc?

I think it's impossible to restrict the groups according to their contents...

adityaghantasala:
One question, does this require that all of those parts have same material ?

jginternational:
@philbucher we can not filter the groups depending on their content, and nothing prevents the user to add entities to the group after the condition is applied.

The only warrant I can give is a validation process in the writing procedures.

jginternational:
@adityaghantasala you mean that the list of materials shown in 'Truss' should be different than the list shown in 'Shells'?
We can work on that direction, but at this moment every app has it's own material database, and it's not filtered by the element type selected.

It's a good idea

jginternational:
Since all the comments are positive, I'll merge this PR soon

Issue #428, opened 2018-07-11
Non assigned boundary conditions should be "null" instead of "0" 


loumalouomega:
This gives problems with plasticity and other problematics

loumalouomega:
Maybe the solution would be differentiate between not set ->null, set but not constrained and set but constrained

jginternational:
Be carefull... 
https://www.youtube.com/watch?v=f7uD6j3Oe0w

This was the origin of the ancient Solid-Structural war

jginternational:
Let's wait to @RiccardoRossi 's and @pooyan-dadvand opinion

philbucher:
+1 for the proposal of @loumalouomega 
It seems like at least for structural its not a war but rather a consenses :+1: 

we are having problems with this over and over again, the latest was @KlausBSautter

pooyan-dadvand:
I have added this to @KratosMultiphysics/technical-committee todo

AlejandroCornejo:
+1 to the proposal! this issue always gives problems (at least for me and some students that use the CL)
Thank you @loumalouomega

rubenzorrilla:
+1. The proposal also makes sense to me. 

But how can this be translated to the GiD GUI? @jginternational would it be possible to add another checkbox as the X, Y and Z constraint that says X, Y and Z do nothing (or something similar)

jginternational:
@rubenzorrilla 
That's already implemented for 3-component (XYZ) parameters. You can use the Enable field in the parameter.
for example:
Enable_X = true
Value_X = 0.0
Enable_Y = false
Enable_Z = true
ByFunction_Z = true
Function_Z = 2*t

(I'm out of the office now so I can't attach a screenshot)

rubenzorrilla:
But eventually this would have to be selected by the user such that if not enabled it the project parameters writing sets null. Am I wrong?

RiccardoRossi:
I am a little lost :-(

is the documentation here wrong?

https://github.com/KratosMultiphysics/Kratos/wiki/How-to-use-Processes-In-the-application-of-BCs

RiccardoRossi:
ping!

pooyan-dadvand:
@KratosMultiphysics/technical-committee would like to know the opinion of @KratosMultiphysics/interface-committee about it as it directly changes the current input.

pooyan-dadvand:
Any comment from @KratosMultiphysics/interface-committee?

[...]

PR #624, opened 2019-07-22
Improve vectorial nodal displacement proposal null #428

jginternational:
Implements #428 
Let's see your impressions by checking some situations in the current GUI, and the equivalent in the new proposal:
### Old GUI
image
You can see that there is a lot of information in the screen. This will generate the following json:
```
{
            "python_module" : "assign_vector_variable_process",
            "kratos_module" : "KratosMultiphysics",
            "Parameters"    : {
                "model_part_name" : "Structure.DISPLACEMENT_Displacement_Auto1",
                "variable_name"   : "DISPLACEMENT",
                "constrained"     : [true,true,true],
                "value"           : [0.0,0.0,0.0],
                "interval"        : [0.0,"End"]
            }
}
```
### New proposal
image
It's a more compact solution (better for small screens) and simplifies the data input. Generates the following json:
```
{
        "constraints_process_list" : [{
            "python_module" : "assign_vector_variable_process",
            "kratos_module" : "KratosMultiphysics",
            "Parameters"    : {
                "model_part_name" : "Structure.DISPLACEMENT_Displacement_Auto1",
                "variable_name"   : "DISPLACEMENT",
                "constrained"     : [true,true,true],
                "value"           : [0.0,0.0,0.0],
                "interval"        : [0.0,"End"]
            }
 }
```
As you can see, it's the same json as the old one

### Old GUI
image
This case generates the following json:
```
{
            "python_module" : "assign_vector_variable_process",
            "kratos_module" : "KratosMultiphysics",
            "Parameters"    : {
                "model_part_name" : "Structure.DISPLACEMENT_Displacement_Auto2",
                "variable_name"   : "DISPLACEMENT",
                "constrained"     : [false,false,true],
                "value"           : [0.0,0.0,"(y/2)*t"],
                "interval"        : [0.0,"End"]
            }
}
```
And the problem here is that the X and Y axis are not set to null, and the json is still valid.

### New proposal
image
Again, you can see a compact design, that generates the following json:
```
{
            "python_module" : "assign_vector_variable_process",
            "kratos_module" : "KratosMultiphysics",
            "Parameters"    : {
                "model_part_name" : "Structure.DISPLACEMENT_Displacement_Auto2",
                "variable_name"   : "DISPLACEMENT",
                "constrained"     : [false,false,true],
                "value"           : [null,null,"(2*z)+t"],
                "interval"        : [0.0,"End"]
            }
}
```
This way, X and Y displacement will be **NULL instead of 0.0**. If I want to set those components to 0.0, I'll need to create another assignation to the same group and select XY as Axis.
image

If you agree, I'll work on this solution and will be applied to all conditions that use the `assign_vector_variable_process` process 

My proposal to (and as a member of) @KratosMultiphysics/interface-committee 
Also ping @KratosMultiphysics/technical-committee


jginternational:
@rubenzorrilla said in https://github.com/KratosMultiphysics/GiDInterface/issues/428#issuecomment-511878495

> From the description above I understand that it is possible to impose more than one direction at the same time only if the value is the same for all the constrained (or initialized) directions.
What I don't like so much is that it is needed to set multiple constraints if the value between the directions to be set is different. The point is that there is no unique manner...
Having said this, I'd be in favour of keeping `XYZ`, `X`, `Y` and `Z` and removing the "double" directions.

jginternational:
ping @KratosMultiphysics/technical-committee

jginternational:
# Meeting 25/07/2019
Attendants: @RiccardoRossi @pooyan-dadvand  @roigcarlo  @rubenzorrilla  @jginternational 

After a short explanation of the pros/cons of this solution:
### pros
- More compact
### cons
- As @RiccardoRossi said, when applying different values in different axis in the same submodelpart, 2 different "vector processes are applied". 

## New proposal by @RiccardoRossi 
For each component `x, y, z` :
- A dropdown that allows the selection of the behaviour:
    - NotApplied: Will set value to `null `and imposed to `false`
    - AppliedByValue: Will set value to `user entry value` and imposed to `false`
    - AppliedByFunction: Will set value to `user entry function` and imposed to `false`
    - ImposedByValue: Will set value to `user entry value` and imposed to `true`
    - ImposedByFunction: Will set value to `user entry function` and imposed to `false`

A simple mockup:
image

As you can see: 
- X is set to value, so lets you insert a value
- Y is set to function, so lets you insert a function
- Z is set to Not apply, so doesn't let you insert anything else

The mockup here would write:

```
{
            "python_module" : "assign_vector_variable_process",
            "kratos_module" : "KratosMultiphysics",
            "Parameters"    : {
                "model_part_name" : "Structure.DISPLACEMENT_Displacement_Auto1",
                "variable_name"   : "DISPLACEMENT",
                "constrained"     : [false,true,false],
                "value"           : [0.0,"3t",null],
                "interval"        : [0.0,"End"]
            }
}
```

rubenzorrilla:
@jginternational which is the status of this?

jginternational:
GUI ready, 
image

Writing pending

jginternational:
@rubenzorrilla you can try it.
The examples are still missing

rubenzorrilla:
> @rubenzorrilla you can try it.
> The examples are still missing

Which kind of examples you need?

jginternational:
I meant the app examples (the buttons in the toolbar) are not updated yet

jginternational:
READY TO MERGE !

PR #627, opened 2019-07-25
Potential 3d example

inigolcanalejo:
Adding a 3d example


jginternational:
As Potential App does not use the time parameters, I've hidden them from the tree

PR #671, opened 2019-11-12
[CDEM] Added procedure to create and assign bonded domains to all volumes

ipouplana:
A procedure was added in the start.tcl file of CDEM app in order to automatically create and assign bonded domains groups to all existing volumes. This was necessary for cases with large number of different bonded domains.

This procedure should not interfere with the standard procedures of the GUI, and can be called from the GiD command line (after loading the CDEM problemtype) just by typing:
`-np- ::CDEM::CreateAndAssign3DBondedGroups`

In the near future, we could also add a graphical icon to call a modified version of this procedure that automatically creates and assigns bonded domains groups only to the selected volumes by the user.


ipouplana:
Ok, I’ll change that

ipouplana:
Ferran just added a button that works for 2D and 3D

kratos.gid/apps/CDEM/examples/BulkGroup.tcl
```diff
@@ -0,0 +1,52 @@
...
+    GiD_Process 'Zoom Frame
+}
+
+proc ::CDEM::examples::CreateAndAssign3DBondedGroups { } {
+
+    set volume_list [GiD_Geometry -v2 list volume]
+
+    for {set i 0} {$i < [llength $volume_list]} {incr i} {
+
+        set volume_id [lindex $volume_list $i]
```
jginternational: Let me change this into a more efficient way

jginternational:
Ready to merge @ipouplana @farrufat-cimne

PR #674, opened 2019-11-15
Create LICENSE.md

jginternational:

Copied from https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/license.txt

Should I add an author name?

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,30 @@
+## Kratos Multiphysics
+BSD 3-Clause License
+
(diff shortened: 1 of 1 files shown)


roigcarlo:
Weren't we using Bsd-4?

philbucher:
I thought so too

RiccardoRossi:
yes we are using the 4-clause ...please keep using that

jginternational:
So, better this way

pooyan-dadvand:
I agree

PR #675, opened 2019-11-21
[@soudah] Fluid non newtonian const. laws with qsvms element

jginternational:
Added 2 constitutive laws forn non newtonian fluids

Changed the element from vms to qsvms


jginternational:
ping @KratosMultiphysics/fluid-dynamics

rubenzorrilla:
We should decide if the reference monolithic element is now the `QSVMS` (this goes in) or if we keep the `VMS` one (this is closed).

If we decide that the  `QSVMS` becomes the reference one, I wouldn't add the non-newtonian laws until we have implemented their 2D version.

rubenzorrilla:
Guys, this PR deals with the addition of the non-newtonian constitutive laws and the setting of the `QSVMS` as reference element in the `FluidDynamicsApplication`.

I can see some changes related with the Wall Shear Stress (WSS) developments, which are out of the scope of the PR and the `FluidDynamicsApplication`. 

Having said this I'd:
- remove the WSS-related changes from this PR
- add the WSS (and other Biomedical magnitudes) to a new GUI derived from the fluid one. Note that these magnitudes make not so much sense when solving non-biomedical problems.

rubenzorrilla (changes requested):
(blocking to avoid accidental merge)

jginternational:
Thanks @rubenzorrilla 

Let @soudah play in this branch. I won't allow the merging until a Bio** Application (derived from fluid dynamics) is created for these changes.

PR #676, opened 2019-11-26
Failure criterion state printed for CDEM

maceligueta:
Allows printing a new result. But I need a bit of help here: the modification of the spd showing the new field should only work for CDEM, not in the general DEM.


jginternational:
If you are using in an app a part of the xml or spd defined in another, and you want to change it without the parent app to notice, there are 2 ways:

- Fully copy: not a valid option for a small change
- Use the event [APPID]::xml::CustomTree to modify the resulting spd

For your case, change only a public name of a static parameter, the second is better.
In `/apps/CDEM/xml/XmlController.tcl` locate the `proc CDEM::xml::CustomTree` and place there something like

`spdAux::SetValueOnTreeItem pn "Show bonds" BondElem ContactMeshOption`

maceligueta:
Thanks for the explanation @jginternational  but how is this
`spdAux::SetValueOnTreeItem pn "Show bonds" BondElem ContactMeshOption`
capable of adding all this information? ->
`<value n="ContactMeshOption"        type="bool" values="true,false"     v="false"           pn="Show bonds"        icon="black1" help="Calculate and show the bonds between particles"/>`

Actually, I don't want to add that item. In this case, I want the bonds to be visible in all DEM guis. But when I added this piece of tree:
`<value n="TypeOfFailure" type="bool" values="true,false" v="false" pn="Type of failure" icon="black1" state="[getStateFromXPathValue {string(../value[@n='ContactMeshOption']/@v)} true]" help="Displays different numbers for different types of failure. 2: tension. 4: shear or combination of stresses. 6: neighbour not found by search. 8: less bonds than minimum"/>`
I want it visible only for CDEM.
Is this possible?

jginternational:
I beg your pardon. I thought the only difference was the circled part, and I didn't see the underlined
image

To add a value to the spd, this is the code:
```tcl

    gid_groups_conds::addF [spdAux::getRoute BondElem] value [list n TypeOfFailure pn "Type of failure" v No values {Yes,No} icon "black1" help "Displays different numbers for different types of failure. 2: tension. 4: shear or combination of stresses. 6: neighbour not found by search. 8: less bonds than minimum"]
    spdAux::SetValueOnTreeItem state {[getStateFromXPathValue {string(../value[@n='ContactMeshOption']/@v)} true]} BondElem TypeOfFailure

```

maceligueta:
I pushed some changes, but they were not tested yet. @farrufat-cimne said he will take care of this.

kratos.gid/kratos.unix.bat
```diff
@@ -20,16 +20,16 @@ fi
 # and maintains OLD_LD_LIBRARY_PATH with previous settings
 # therefore, we use the OLD_LD_LIBRARY_PATH and prepend the path to the kratos libs
 if [ "$OLD_LD_LIBRARY_PATH" != "" ]; then
-    export LD_LIBRARY_PATH="$3/exec/Kratos":"$3/exec/Kratos/libs":$OLD_LD_LIBRARY_PATH
+    export LD_LIBRARY_PATH="$3/exec/Kratos/bin/Release":"$3/exec/Kratos/bin/Release/libs":$OLD_LD_LIBRARY_PATH
```
jginternational: 👎

maceligueta: ouch! That slipped in!! Will remove it immediately

maceligueta:
Thanks for pointing out @jginternational . I think I undid it.

maceligueta:
I recovered the Parallel configuration, which works in my machine. What type of problem was it giving on your pc, @farrufat-cimne ?
I also recovered Gravity, otherwise we can't modify its values. Please @farrufat-cimne confirm I am not making a mistake.

maceligueta:
One more question @farrufat-cimne :+1: it seems that we are removing fields from the tree here
https://github.com/KratosMultiphysics/GiDInterface/blob/68dc40c0b3d37867ea6db7532d77454558daee80/kratos.gid/apps/DEM/xml/XmlController.tcl#L31
When the procedure should be the opposite, no? I thought the CDEM would increment the fields.
@jginternational , if we reverse the operation, adding fields in CDEM instead of removing in DEM, how can we proceed? Here we are adding one field to the tree
https://github.com/KratosMultiphysics/GiDInterface/blob/16cff5750026af93ea1539546259ec6946dfebca/kratos.gid/apps/CDEM/xml/XmlController.tcl#L34
But I guess adding new conditions is different...

jginternational:
Solved by phone.
Conditions that must only appear in the child app should not exist in the parent app, They must be loaded in a xml file in the child one

maceligueta:
I think this is ready. I tested examples in DEM and CDEM and they run fine

maceligueta:
@jginternational approve if you think everything is ok

maceligueta:
I am not authorized to merge, even if it's approved...

jginternational:
We are on _Christmas_ ... on release period

PR #680, opened 2019-11-28
Feature mpm 2D example

jginternational:
@VeronikaSinger Check the 2D example

It's not working with the current python scripts


jginternational:
You can test this branch and test the example button:
image

@antonialarese @VeronikaSinger 

[mpm_mesh_examples.zip](https://github.com/KratosMultiphysics/GiDInterface/files/4129401/mpm_mesh_examples.zip)

In this folder you can find the same example, with different mesh sizes.
The 0.2 case works fine, but the 0.1 case stops for no reason before "the sand contacts the floor"

VeronikaSinger:
Hi Javi,

I checked the examples in the .zip folder. For the smaller mesh, the time step is too large. If it is reduced  to 0,0005 than it works. 
But when I run both examples you sent an error occurs as soon as the particles get in contact with the slip condition. The problem is the initial definition of the normals. If I swap the normals accordingly both examples run until the end. 
Screenshot from 2020-02-06 10-38-38
I also checked the example in GiD itself. If you have a triangular mesh it is not possible to discretize it with 10 particles per element. Use either 3, 6 or 12 particles per element. But this will not throw you an error. But also for this case you have to swap the normals.
A "nice to have feature" would be, if in the ProjectParameter.json the gauss_point_results for MP_Displacement, MP_Velocity would be printed by default.

Issue #683, opened 2019-12-03
DEM wall for linear line elements


KlausBSautter:
Hi, I think I was using an old version of the Interface for the DEM App. Anyway here is my problem:

With the version, I was using I could create a DEM-FEM wall condition for a group with line elements and it created 
> RigidFace3D2N

 with the current interface, it creates

> RigidFace3D3N

automatically. How can I change this behavior? 

Klaus

maceligueta:
Hi @KlausBSautter. Thanks for reporting. Just one question (due to my bad memory): in the old GUI, in 3D we allowed Line elements and they were called RigidFace3D2N?
This is just to state clearly what we want from the new GUI.
Right now @farrufat-cimne and myself are changing parts of the GUI, but the real expert is @jginternational , so I will keep him in the loop, understanding that this issue is quite urgent because it is blocking your work somehow.

KlausBSautter:
@maceligueta yes I need the line elements as contact conditions for the protection nets. In the GUI you didn't need to specify the element, you just assigned a group to the dem-fem wall condition and if it contained only lines it created RigidFace3D2N.

jginternational:
Hi @KlausBSautter 

Correct me if I'm wrong.
You are using DEM application in 3D and you want to apply the DEM-FEM-WALL condition to a group containing lines, right?

KlausBSautter:
yes, exactly. Somehow this was working before

jginternational:
Is this done?

KlausBSautter:
yes, I guess. It works now. Thanks!

PR #684, opened 2019-12-03
Feature/dem allow line in demfemwalls

jginternational:
Implements #683 

@KratosMultiphysics/dem


maceligueta:
Hi @jginternational , will this work with added meshes that are imported after meshing? We had a problem a few weeks back because the loop was being done ONLY over groups which were not empty in the GiD Geometry Mode (it was checking if there was lines or surfaces as geometric entities). But we sometimes leave empty groups, with conditions applied, we then mesh and after that we import some STL.

jginternational:
@maceligueta I understand. Let me check it.

Anyway, @escolano recommends to import those stl meshes, transform them into geometrical entities via `Geometry > Create > Geometry from Mesh`

PR #687, opened 2019-12-11
Meshcriteria is resetted on all surfaces of the model

maceligueta:
The problem was the following:
1- Group created and assigned to condition DEM-FEM-Wall
2- Mesh (MeshCriteria Mesh is applied to the surfaces in the group)
3- You unassign some surfaces from that group
4- Mesh -> BUG: some surfaces are being meshed that should not be. Nobody was resetting the MeshCriteria Mesh.
I don't see how I can fix this bug in a different way. I know that this may interfere with other applications coupled with DEM.


jginternational:
I think it's ready @maceligueta

Issue #690, opened 2019-12-18
[dem] Dem Parts in 2D allow assigning to Volumes, when they should not


maceligueta:
This is general, but should be different in 2D/3D
https://github.com/KratosMultiphysics/GiDInterface/blob/a1d5208333049fdacb6369f6e1bcb3885ff5cb76/kratos.gid/apps/DEM/xml/Parts.spd#L3

jginternational:
You'll need something like this:
```xml
 <condition n="Parts" pn="Parts" ov="[GetDEMPartsOv]" ovm="node,element" icon="shells16" help="Select your group" un="DEMParts"> 
```

and, in the procs.spd file, implement the function that returns "point,line,surface" in 2d and "point,line,surface,volume" in 3D

maceligueta:
thanks! It is working. Will start a new PR for it.

jginternational:
If you want it to be included in the release, be fast. In 1 hour it'll be out of the oven!

maceligueta:
Well, several other bugs will be packed then. We spent the day reporting bugs related to DEM

PR #694, opened 2019-12-18
[dem] Parts can select different types of entities depending on number of dimensions

maceligueta:
Solves #690


jginternational:
Tested

Issue #677, opened 2019-11-26
[dem] Custom submodelpart not printing anything


maceligueta:
I just created an example with the standard DEM and tried to create a custom submodelpart with with one sphere, destination DEM mdpa. Nothing was printed there.
@farrufat-cimne 
Referencing also @jginternational in case you have time to look at it, too.

jginternational:
It is commented here:

https://github.com/KratosMultiphysics/GiDInterface/blob/afe26b84e089b90f770e2e71185eb9e597e431e7/kratos.gid/apps/DEM/write/writeMDPA_Parts.tcl#L33

PR #697, opened 2019-12-23
Hotfix dem customsmp

jginternational:
This implements #677

It fails because, as it prints spheres in the DEM.mdpa, those spheres need properties, and fail if no const law is defined inside it....

Please test it and decide about it @KratosMultiphysics/dem


jginternational:
Notes:

- Only spheres existing in Parts can appear here

farrufat-cimne:
Looks good, I'm going to test it and see if everything is good.
@maceligueta

PR #708, opened 2020-01-23
Part names must be unique to transform

jginternational:


jginternational:
This will break some models in transform, but it'll work 'forever' (hope)

PR #709, opened 2020-01-23
Conditions parameters - missing clausule

### LINKED ISSUE #703: [DEM] Bug in tree. Fields appear when they should not (and vice versa)
ALL Options in DEM-FEM wall condition appear even if type of motion is set to None

jginternational:
@farrufat-cimne test to fix #703

### DIFF
--- a/kratos.gid/scripts/Controllers/CommonProcs.tcl
+++ b/kratos.gid/scripts/Controllers/CommonProcs.tcl
@@ -782,11 +782,13 @@ proc spdAux::ProcOkNewCondition {domNode args} {
 
 proc spdAux::ProcConditionParameterState {domNode args} {
     set ret ""
+    # Current parameter name
     set param_name [get_domnode_attribute $domNode n]
+    
+    # Condition
     set cond_node [$domNode parent]
     if {[$cond_node nodeName] eq "group"} {set cond_node [$cond_node parent]}
     set cond_name [get_domnode_attribute $cond_node n]
-    
     set cond [Model::getCondition $cond_name]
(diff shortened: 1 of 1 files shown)


kratos.gid/scripts/Controllers/CommonProcs.tcl
```diff
@@ -805,12 +807,14 @@ proc spdAux::ProcConditionParameterState {domNode args} {
...
-            set realV [get_domnode_attribute [$domNode selectNodes "../value\[@n='$depN'\]"] v]
-            if {$depV ne $realV} {set ret hidden}
+            set parent_dependency_node [$domNode selectNodes "../value\[@n='$depN'\]"]
+            set parent_dep_state [get_domnode_attribute $parent_dependency_node state]
+            set realV [get_domnode_attribute $parent_dependency_node v]
+            if {$realV ni [split $realV ','] || $parent_dep_state eq "hidden"} {set ret hidden}
         }
     }
     if {$ret eq  ""} { set ret normal }
-    return normal
```
jginternational: image

farrufat-cimne:
Gonna test it. 

thx

farrufat-cimne:
@jginternational it looks good, but there is a minor tree refresh issue that is still persistent.
If the field is initially set to NO, child fields are shown when they should not.
If the parent field changes to YES,  child fields are shown as they should.
If the field is switched to NO, child fields are hiddden as expected.

jginternational:
Right, thats a known issue. Should be fixes on GiD side

PR #714, opened 2020-02-05
[DEM]graph changes, custom smp update and additional cluster features

farrufat-cimne:
The option to generate graphs with DEM is now set as condition, both inside one of the existing conditions on DEM (as a flag in initial condition // boundary condition) and as separate condition (Define graphs using particles)
I would remove the last one (Define graphs using particles) since it's not clear that a user would decide to generate a graph without previously assign another condition, for example a boundary condition on the particles. 
Typical example would be a group of "fixed" particles used to measure the force against them.

_______________________________
Also, I was not able to run simulations when defining custom dem mdpa with the code updated in:
[2b74bf37e35275d50eb3ca1513f7967cf947c4f2 ](https://github.com/KratosMultiphysics/GiDInterface/commit/2b74bf37e35275d50eb3ca1513f7967cf947c4f2#diff-949c0e22b24cacb6d47151bd01155f06)

```
`proc DEM::write::WriteCustomDEMSmp { } {
    foreach group [GetDEMGroupsCustomSubmodelpart] {
        set groupid [write::GetWriteGroupName [$group @n]]

        # TODO: Missing write properties for Custom Submodelparts

        # Nodes are previously printed
        # Print elements and connectivities
        set elem [write::getValueByNode [$group selectNodes ".//value\[@n='Element']"] ]
        write::writeGroupElementConnectivities $group $elem

        DEM::write::writeSphereRadiusOnGroup $group

        write::writeGroupSubModelPart DEM-CustomSmp $groupid Elements
    }
}`
```

but it works with the previous proc still defined in the CDEM version:

`
```
proc CDEM::write::WriteCustomDEMSmp { } {
    set xp1 "[spdAux::getRoute [GetAttribute conditions_un]]/condition\[@n = 'DEM-CustomSmp'\]/group"
    foreach group [[customlib::GetBaseRoot] selectNodes $xp1] {
	set groupid [$group @n]
	set destination_mdpa [write::getValueByNode [$group selectNodes "./value\[@n='WhatMdpa'\]"]]
	if {$destination_mdpa == "DEM"} {

	    write::WriteString  "Begin SubModelPart $groupid \/\/ Custom SubModelPart. Group name: $groupid"
	    write::WriteString  "Begin SubModelPartData"
	    write::WriteString  "End SubModelPartData"
	    write::WriteString  "Begin SubModelPartNodes"
	    GiD_WriteCalculationFile nodes -sorted [dict create [write::GetWriteGroupName $groupid] [subst "%10i\n"]]
	    write::WriteString  "End SubModelPartNodes"
	    write::WriteString  "End SubModelPart"
	    write::WriteString  ""
	}
    }
}`
```

Added several other bugfixes.


farrufat-cimne:
There is a problem when assigning default names to submodelpart containing spaces:

`Begin SubModelPart Define custom DEM Submp Auto1 // Custom SubModelPart. Group name: Define custom DEM Submp Auto1`
changes to:
`Begin SubModelPart domNode0xa042370 // Custom SubModelPart. Group name: Define custom DEM Submp Auto1
`

jginternational:
> There is a problem when assigning default names to submodelpart containing spaces:
> 
> `Begin SubModelPart Define custom DEM Submp Auto1 // Custom SubModelPart. Group name: Define custom DEM Submp Auto1`
> changes to:
> `Begin SubModelPart domNode0xa042370 // Custom SubModelPart. Group name: Define custom DEM Submp Auto1 `

Dont use domnode object as name. You shoud call the functions


```tcl
set group_id [$group @n]
set group_raw [write::GetWriteGroupName $group_id]
set good_name [write::transformGroupName $group_raw]
set submodelpart_id "${cid}_${good_name}"
```

Explanation:
- write::GetWriteGroupName -> This is necessary only if the condition has intervals, because we use group parent/child scheme to implement intervals, and we don't want to print the interval information.
for example: "Inlet//Total" and "Inlet//custom_interval1" will both return "Inlet"

- write::transformGroupName -> This tool removes invalid chars like blankspeces or tabs and places "_".
for example: "group name" will be "group_name" and "parent group//child group" will be "parent_group/child_group"

- The last command is there because we create the submodelpart names including the condition id and the group id, so we can easily recognize which submodelpart comes from each part of the tree

farrufat-cimne:
The custom cluster feature could be improved with some tweaks.

farrufat-cimne:
Examples were not running but it was not a problem of the interface.

maceligueta:
Hi @farrufat-cimne , I think you should change the title of this PR to reflect all the issues you are tackling.

jginternational:
I see you are using parameters with type = file.

If you write here how do you want to use that I can help with the write procs, because the only app who is using it at this moment is @KratosMultiphysics/dam and it's not public yet

joaquinirazabal:
Hi, @jginternational about the @KratosMultiphysics/dam application it can be made public when you want/can. If you need anything else from @djvicente or me tell us :)

farrufat-cimne:
The simulation seems to be working fine. 
When selecting cluster from file and checking the "Save file into model" the cluster_file is copied from its original location to the working folder. But, when the simulation is launched, a warning appears trying to copy (again) the same clusterfile.

farrufat-cimne:
fixed the warnings and checked for both DEM and CDEM ui:
- non cluster inlet
- predefined cluster inlet
- customized cluster inlet via "file"

all combinations seems to be working fine now. No warnings detected.

Issue #713, opened 2020-01-31
PFEM Fluid improvements


jginternational:
- Write materials json file
    - All properties -> 0
    - Loop Bodies

- Add "Thermally coupled" in Physics
  - Parts properties: (Add from ConvDiff)
    - CONDUCTIVITY
    - SPECIFIC HEAT
  - Boundary Conditions:
    - TEMPERATURE (ConvDiff)
  - Solver Settings: (ConvDiff) bring
    - Analysis type
    - Strategy parameters
    - Linear Solver

- Delete BC:
  - Displacement
  - Acceleration

@MZecchetto

jginternational:
```
{
    "problem_data"         : {
        "problem_name"   : "coupling_test_1_mesh_size_04",
        "start_time"     : 0.0,
        "end_time"       : 15.0,
        "echo_level"     : 0,
        "parallel_type"  : "OpenMP",
        "threads"        : 8,
        "gravity_vector" : [0.0,-9.81,0.0]
    },
    "solver_settings"      : {
        "solver_type"       : "pfem_fluid_thermally_coupled_solver",
        "domain_size"                        : 2,
        "time_stepping"                      : {
            "automatic_time_step" : true,
            "time_step"           : 0.002
        },
        "fluid_solver_settings": {
            "solver_type"                        : "pfem_fluid_solver",
            "model_part_name"                    : "PfemFluidModelPart",
            "physics_type"                       : "fsi",
            "domain_size"                        : 2,
            "model_import_settings"              : {
                "input_type"     : "mdpa",
                "input_filename" : "coupling_test_1_mesh_size_04"
            },
            "time_stepping"                      : {
                "automatic_time_step" : true,
                "time_step"           : 0.002
            },
            "maximum_pressure_iterations"        : 7,
            "velocity_tolerance"                 : 1e-5,
            "pressure_tolerance"                 : 1e-5,
            "echo_level"                         : 1,
            "velocity_linear_solver_settings"    : {
                "solver_type"         : "bicgstab",
                "max_iteration"       : 5000,
                "tolerance"           : 1e-9,
                "preconditioner_type" : "ilu0",
                "scaling"             : false
            },
            "pressure_linear_solver_settings"    : {
                "solver_type"         : "bicgstab",
                "max_iteration"       : 5000,
                "tolerance"           : 1e-9,
                "preconditioner_type" : "ilu0",
                "scaling"             : false
            },
            "bodies_list"                        : [{
[...]

PR #717, opened 2020-02-14
Pfem/materials file

jginternational:
Materials file written instead of mdpa properties

implements first point of #713 
implements KratosMultiphysics/Kratos/pull/6258

### DIFF
--- a/kratos.gid/apps/PfemFluid/write/write.tcl
+++ b/kratos.gid/apps/PfemFluid/write/write.tcl
@@ -22,6 +22,9 @@ proc PfemFluid::write::Init { } {
 
 # Model Part Blocks
 proc PfemFluid::write::writeModelPartEvent { } {
+    # Init data
+    write::initWriteConfiguration [GetAttributes]
+    
     set parts_un_list [GetPartsUN]
     foreach part_un $parts_un_list {
         write::initWriteData $part_un "PFEMFLUID_Materials"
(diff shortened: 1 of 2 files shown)


MZecchetto:
Hi @jginternational ,
first comment. we need to add the following to the `solver_settings` in the `ProjectParameters.json`
`"material_import_settings"           : {
            "materials_filename" : "PFEMFluidMaterials.json"
        },`

Second comment. we need to remove the constitutive laws in the `PFEMFluidMaterials.json` as they are not supported by the PFEMFluidApp. I would comment the script instead of delete it, since it can be useful in the near future.

MZecchetto:
Third comment. Is possible to have a property for each group of elements? In particular also for each rigid boundary group. In this case they are kept separated in the GiD post process. For the 2D example this means add to the PFEMFluidMaterials:

`{        "model_part_name" : "PfemFluidModelPart.Parts_Rigid_Walls",

        "properties_id"   : 2,

        "Material"        : {
            "Variables"        : {},
            "Tables"           : {}
        }
    }`

jginternational:
@MZecchetto Points 1 and 2 are done. 
Point 3 is a bit difficult and I don't know if setting "empty properties" in the materials json is the proper way

MZecchetto:
With these last two commits everything works fine. 
Adding "empty properties" works for the pfem app, also with the material.json. As far as I understood, the GiD post process uses the property Id to assign the elements to the different layers. Are there better ways to do so?

PR #735, opened 2020-04-09
[All] Linear solvers clean up

rubenzorrilla:
Required changes according to https://github.com/KratosMultiphysics/Kratos/pull/6741.

@jginternational note that the `ExternalSolversApplication` is being deprecated in favour of the `EigenSolversApplication`. We should start packaging the second one instead of the first one.

Another question I have is why the PFEM applications are not using the common `solvers.xml`. As far as I see they have exactly the same solvers.

### DIFF
--- a/kratos.gid/apps/Common/xml/Solvers.xml
+++ b/kratos.gid/apps/Common/xml/Solvers.xml
@@ -7,22 +7,12 @@
         </inputs>
     </solver>
 
-    <solver n="ExternalSolversApplication.super_lu" pn="SuperLU" SolverType="Standard" Parallelism="OpenMP">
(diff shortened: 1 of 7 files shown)


jginternational:
> Required changes according to [KratosMultiphysics/Kratos#6741](https://github.com/KratosMultiphysics/Kratos/pull/6741).
> 
> @jginternational note that the `ExternalSolversApplication` is being deprecated in favour of the `EigenSolversApplication`. We should start packaging the second one instead of the first one.

Perfect!

> 
> Another question I have is why the PFEM applications are not using the common `solvers.xml`. As far as I see they have exactly the same solvers.

You can guess 👀

PR #738, opened 2020-04-30
Feature/add pfem flow contlaws

jginternational:
Implements #713


MZecchetto:
Hi @jginternational, it works for me, thanks!
I just added the constitutive name in the material.json again

PR #739, opened 2020-04-30
Feature/add pfem fluid thermic

jginternational:


jginternational:
Implements #713

AFranci:
Thanks Javi, 

we would need as soon as possible the rest of the developments of issue #713 on the thermal coupled solver.
At this moment, we are in a dead point... we cannot go ahead with the development in kratos and the interface has some thermal elements that cannot be used directly.
Thank you for your cooperation!

PR #748, opened 2020-07-01
more events called

maceligueta:


maceligueta:
@jginternational If it looks OK you can merge directly.

jginternational:
There is an extra thing to be done if you want it to run in gids older than version 14.1.7d (or official 14 versions)

PR #756, opened 2020-09-09
Hotfix/dem save spheres in group

### LINKED ISSUE #755: [dem] Allow/Disallow element type of a dem groupd has changed after saving and loading a case
Process to reproduce bug:

- Load Gid 15 and the Kratos Gid Interface 
- Choose DEM Cohesive, 3D
- Generate a volume and assign it to DEM parts
- Mesh
- Check that both the spheres and the nodes belong to the group (Yes). Check that all element types are allowed (Yes)
- Save the case
- Close GiD
- Open the case
- Check that the spheres are no longer in the group. Check that now only some element types are allowed, Sphere is among the Disallowed ones now.

jginternational:
Fixes #755 

As suspected, it was a GiD bug in the lasts versions (most 14th and first 15th). 

This patch will apply to all GiD versions until 15.0.0 (included), so in the next version the bug will no longer exist 🥇 

As you can see, all the apps depending on the DEM were experiencing this issue, so it's fixed for all of them.

Take a look @maceligueta 

PD: this branch comes from another one, which is about to merge. The only commits important for you are the last 3

### DIFF
--- a/kratos.gid/apps/CDEM/start.tcl
+++ b/kratos.gid/apps/CDEM/start.tcl
@@ -71,4 +71,8 @@ proc ::CDEM::CustomToolbarItems { } {
     Kratos::ToolbarAddItem "Bulk grouping" [file join $dir images stone.png] [list -np- ::CDEM::xml::BulkGroup] [= "Plugin\nBulk grouping"]
 }
 
+proc ::CDEM::AfterSaveModel {filespd} {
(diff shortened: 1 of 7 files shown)


maceligueta:
oops! I merged. I tested it and works perfectly. After that I realized that this PR contains other stuff. Should I roll back the merge?

jginternational:
It's fine!

PR #758, opened 2020-09-22
Hide end calculation window

jginternational:
This PR allows the apps to remove the windows after the calculation process
And sets the DEM app as example:

Caution:
- Works with 15.1.0d and later
- Also removes the error window if the calculation process went wrong.

@maceligueta @farrufat-cimne

### DIFF
--- a/kratos.gid/apps/DEM/start.tcl
+++ b/kratos.gid/apps/DEM/start.tcl
@@ -177,6 +177,13 @@ proc ::DEM::AfterSaveModel { filespd } {
     }
 }
 
+proc ::DEM::AfterRunCalculation {basename dir problemtypedir where error errorfilename} {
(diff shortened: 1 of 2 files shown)


maceligueta:
Hi @jginternational  and @farrufat-cimne . We did not need this for the DEM. It was intended for an independent problemtype.

I am OK with it if it's useful in the future or for other apps, however, as I said, **we do not currently need this for the DEM**.

jginternational:
Ok, it is commented for the DEM. It was just an example for @farrufat-cimne .

As you can see, once implemented the event GiD_Event_AfterRunCalculation , if you don't want that window, just return -cancel-

maceligueta:
Great then!

farrufat-cimne:
Thx Javi

PR #762, opened 2020-10-06
bugfix DEM 2d Circle mesher

farrufat-cimne:
temporal fix for example 2D. Assigned Rball instead of Granular until the mesher is fixed


maceligueta:
Hi @farrufat-cimne , we should first save in a variable the mesher that was chosen by the user, and restore the option at the end to that value. Otherwise, the user will be shocked that the chosen mesher changes randomly.

jginternational (approved):
See how can we access or change GiD preferences using GiD_Set command

farrufat-cimne:
Thanks Javi!

PR #764, opened 2020-10-19
Update fluiddem interface

jgonzalezusua:
Updating FluidDEM interface to be able to select _two-way_ problem parameters and fixing bugs in master branch.

### DIFF
--- a/kratos.gid/apps/DEM/write/writeProjectParameters.tcl
+++ b/kratos.gid/apps/DEM/write/writeProjectParameters.tcl
@@ -69,8 +69,8 @@ proc DEM::write::getParametersDict { } {
 
     dict set strategy_parameters_dict "RemoveBallsInitiallyTouchingWalls"   [write::getValue AdvOptions RemoveParticlesInWalls]
     dict set strategy_parameters_dict "strategy"                            $dem_strategy
-    dict set project_parameters_dict  "solver_settings"                      $strategy_parameters_dict
 
+    dict set project_parameters_dict "solver_settings"                      $strategy_parameters_dict
     dict set project_parameters_dict "VirtualMassCoefficient"               [write::getValue AdvOptions VirtualMassCoef]
     dict set project_parameters_dict "RollingFrictionOption"                [write::getValue AdvOptions RollingFriction]
     dict set project_parameters_dict "GlobalDamping"                        [write::getValue AdvOptions GlobalDamping]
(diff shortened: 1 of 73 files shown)


GuillermoCasas:
I would approve but I prefer @maceligueta to please take a quick look at the new features added to the DEM interface. First, search tolerance is added (I believe), second, the phantom walls are added. Both things are important for the regular DEM interface.

kratos.gid/apps/DEM/write/writeProjectParameters.tcl
```diff
@@ -69,7 +69,7 @@ proc DEM::write::getParametersDict { } {
 
     dict set strategy_parameters_dict "RemoveBallsInitiallyTouchingWalls"   [write::getValue AdvOptions RemoveParticlesInWalls]
     dict set strategy_parameters_dict "strategy"                            $dem_strategy
-    dict set project_parameters_dict "solver_settings"                      $strategy_parameters_dict
+    dict set project_parameters_dict  "solver_settings"                      $strategy_parameters_dict
```
GuillermoCasas, line 72: extra space

kratos.gid/apps/FluidDEM/xml/XmlController.tcl
```diff
@@ -78,7 +80,7 @@ proc FluidDEM::xml::CustomTree { args } {
 
     # set result_node [$root selectNodes "[spdAux::getRoute FLResults]/container\[@n = 'GiDOptions'\]"]
 	# if { $result_node ne "" } {$result_node delete}
-    spdAux::SetValueOnTreeItem state disabled FLSolStrat
+    spdAux::SetValueOnTreeItem state enabled FLSolStrat
```
maceligueta, line 83: If we were disabling this item maybe it's not necessary to enable it explicitly and this line could be removed.

kratos.gid/apps/FluidDEM/xml/Main.spd
```diff
@@ -5,6 +5,7 @@
     <include n="AnalysisType"        active="1" path="apps/Fluid/xml/AnalysisType.spd"/>
     <include n="FluidParts"          active="1" path="apps/Fluid/xml/Parts.spd"/>
     <include n="InitialConditions"   active="1" path="apps/Fluid/xml/InitialConditions.spd"/>
+    <include n="ManufacturedConditions" active="1" path="apps/FluidDEM/xml/ManufacturedConditions.spd"/>
```
maceligueta, line 8: Are you sure this is of general interest? I mean, if it's for testing, it should be in the tests, but not in the GUI. No normal user should be interested in using ManufacturedConditions.

jgonzalezusua: I think could be useful. If you know the exact solution of a variable you can impose it by using that condition. In the tree is set as "Exact Conditions"

jgonzalezusua:
At the moment, the ManufacturedConditions are an internal feature we are using and, if it is not of general interest, we can remove it in the future. @maceligueta, can you approve the PR?

maceligueta:
I am approving, but it would be better if this feature is only available to developers (when that option is chosen) as we have spoken in person.

maceligueta (approved):
Approving to speed up the process but switch to developer mode pending if it is possible

maceligueta (approved):

PR #767, opened 2020-10-30
[Fluid] Export new elements to GUI

rubenzorrilla:
This is what I told you. I manage to add the new elements and the custom function to write the `formulation` field of the `solver_settings`.

What I didn't manage to do is to implement the element type filtering according to the time scheme selection.


jginternational:
@rubenzorrilla you can try it now. 
Do you know the joke... that guy that brings his car to a mechanical workshop and after a few moments, the mechanic touches a screw and says, "Ok, it's all done, It will cost 10.000$" ?  -> Same happends here. It's just a little screw that caused the malfunction

rubenzorrilla:
Now the issue, which I think we cannot solve, is to filter the `dynamic_tau` in the strategy settings according to the chosen element.

jginternational:
Wich are the elements that enable that dynamic_tau?

rubenzorrilla:
> Wich are the elements that enable that dynamic_tau?

All the elements except the `DVMS2D` and `DVMS3D`.

jginternational:
Ya estaría

PR #769, opened 2020-11-11
Hotfix -> Gauss point result hidden in some apps like Structural

jginternational:
The element results container was hidden due to a change in the way the Structural app shows it's elements.

Now this is what user sees when selecting a Solid part, with Small displacement element
image


jginternational:
What we saw this morning @RiccardoRossi @rubenzorrilla

PR #772, opened 2020-11-17
Feature/improved examples window

jginternational:
New examples window
image
image

Every app defines it's examples, and they are displayed in a common window. When this window is called from an app, the examples are filtered, so you can only see those involving that app.

The images have been uploaded by me, but feel tree to change them. You'll do it better than me. Just create a 100x100px image and push it to your app images folder with the proper name.

This improvement is important for the user experience, but also for the testing tool (mainly I've created this to make my own life easier when testing.

You can see this window in action when clicking the "Examples" button on the main screen, or the examples icon on the left toolbar when you are on an application.

Just to inform @KratosMultiphysics/all that this will be included in the next release, so if you want to add examples, or change the existing ones, please check it!


jginternational:
With this fix, I'm pushing it to master

PR #781, opened 2020-12-01
Feature/increase logs

jginternational:
Improve the log system


jginternational:
Added to preferences the decision for logs:
- Don't log
- Local log -> Logs stored in the model, so support can trace what happened to the model
- Send anonymous usage data and help us to improve (not ready)

image

Anyone can add anything to the log using
Kratos::Log "message to log"

PR #793, opened 2021-01-14
Allow eigensolver to MPI

### LINKED ISSUE #776: Writing error with MPI parallel type
When selecting MPI parallel type a writing error pops up. This happened to me in the FSI application.

jginternational:
Fixes #776


rubenzorrilla:
Not valid as the eigensolver doesn't exist in MPI.

PR #800, opened 2021-02-26
Temperature dependency properties

rlrangel:
Adding temperature dependency tables for more material properties and a few adjustments.


AlejandroCornejo (approved):
ok for me

PR #807, opened 2021-05-06
Problem type of GeoMechanicsApplication

Vahid-Galavi:
At the moment, GeoMechanicsApplication has its own problem type.  All features of GeoMechanicsApplication are available in this problem type.


jginternational (changes requested):
Hi @Vahid-Galavi 

I can see that this is a 'classic' problemtype. We are no longer using this system because it does not allow integration of your application in the Kratos ecosystem.

I can help you with the migration. I've been looking at your code and it will be easy since you already migrated to submodelparts and process in json.

Vahid-Galavi:
> Hi @Vahid-Galavi
> 
> I can see that this is a 'classic' problemtype. We are no longer using this system because it does not allow integration of your application in the Kratos ecosystem.
> 
> I can help you with the migration. I've been looking at your code and it will be easy since you already migrated to submodelparts and process in json.

Thanks @jginternational! It would be great to transform this problemtype to more generic one that you have been developing. Please let me know how we should proceed.

jginternational:
@Vahid-Galavi  I think the next step is:

- I'll prepare the skeleton of the application.
- I'll fill it with some of the elements, constitutive laws and conditions...
- I'll prepare for the basic write proc using the common code.

After that
- You'll be able to add the remaining elements, constitutive laws, conditions... in an easy way (xml)
- You'll be able to change the writing process to customize. Take into account that if your mdpa and json files are following the standard, the write process is almost automatic. You won't need to print every single process or element.


It will also help if you provide a **basic example** (the minimum example that runs) and a **complex example** (trying to use most of the features, so I can see them working).

PR #813, opened 2021-06-03
Dem/mat rel window

jginternational:
You'll be able to check if there's any missing material relation. Just click on the icon. 
image

image


maceligueta:
This is great @jginternational . Thanks!

maceligueta (approved):

PR #818, opened 2021-06-29
[FluidDem] Material relations

jginternational:
Update Fluid DEM application to Material relations changes in DEM


jginternational:
@jgonzalezusua Can we merge this branch? I'm trying to run the FluidDem3D example and it does not work. I get a segmentation fault running the master branch of  Kratos

jgonzalezusua:
Looks like the interface works. About the error in the FluidDEM example, it could be caused by changes on Kratos' master branch and maybe it needs to be updated. I think it is not related to the interface.

PR #823, opened 2021-08-01
SwimmingDEM add permeability

jginternational:
Adds permeability to Fluid Materials in Swimming DEM.
@KratosMultiphysics/dem 

This is a first approach to Matrix input

GiD team is working on a Matrix input for CustomLib, so in the next GiD version we will have an improved GUI for this kind of inputs.
Probably in September 2021


jgonzalezusua:
I have seen that in the FluidMaterials.json the permeability is set as six different values instead of a matrix. By the way, we need that the problemtype writes the matrix in 2D cases as well, for this, it is only necessary to provide three elements (A11, A22 and A12).

jgonzalezusua:
ping @jginternational

jginternational:
We have developed this at the GiD side.
Screenshot 2021-08-04 at 11 41 06

We will be able to add nxm matrix, so instead of doing an intermediate development, I'll do it directly using this new feature.
Pros: It will be a definitive development, homogeneous with the other kratos apps, and the rest of GiD Problemtypes
Cons: It will be ready in the next GiD Version (Developer and official)

jgonzalezusua:
Ok, the release of the next version will be next month, right?

jginternational:
It will be ready in the next GiD version

```<parameter n="PERMEABILITY" pn="Permeability" v="1e+89,0.0,0.0,0.0,1e+89,0.0,0.0,0.0,1e+89" type="matrix" rows="3" cols="3" />```

image

It will also work for 2D, but it is not enabled the app Swimming DEM 2D

**Comment by jginternational** [comment]
The only missing thing is that matrix is being printed like this:
```
"Variables"        : {
                "PERMEABILITY"      : "{1e+89 0.0 0.0} {0.0 1e+89 0.0} {0.0 0.0 1e+89}",
                "DENSITY"           : 1.0,
                "DYNAMIC_VISCOSITY" : 0.002
            }
```

Instead of like this:
```
"Variables"        : {
                "PERMEABILITY"      : [[1e+89, 0.0, 0.0], [0.0, 1e+89, 0.0], [0.0, 0.0, 1e+89]],
                "DENSITY"           : 1.0,
                "DYNAMIC_VISCOSITY" : 0.002
            }
```

jginternational:
Merging this  @jgonzalezusua

PR #828, opened 2021-08-13
CHT nodes in parts (and another apps)

jginternational:


rubenzorrilla (approved):
Fine if it fixes the problem. Thanks!

PR #835, opened 2021-09-19
DEM example. I found this old fix I had in my pc. Not sure if it is correct at the moment

maceligueta:
I created this PR in order not to lose the changes or forget the problem.


jginternational:
El cambio es un espacio en blanco!

maceligueta:
It wasn't just a space in my head XD. I  deactivated 'also lower entities' when configuring the meshing options of 3 types of objects. omg. I'll find that someday... or not.

PR #854, opened 2021-11-16
Condition only existing groups

jginternational:
This PR enables the new GiD feature (15.1.4d) to configure conditions:

When double click on a condition, dont allow the creation of new groups. Execute a tcl proc to return a group list, so the user can select from that list. Useful in the new @KratosMultiphysics/dem  app

Copy and pase this sections to see the changes (I don't add them to the PR because we dont want them to exist yet)
- file `kratos.gid/apps/DEM/xml/Conditions.xml` line 21: 
```xml
 <ConditionItem n="DEM-FEM-Wall2D" pn="Rigid Walls" Interval="False" 
ImplementedInApplication="DEMApplication" MinimumKratosVersion="9000" 
ProductionReady="ProductionReady" WorkingSpaceDimension="2D" 
LocalSpaceDimension="1" ElementType="Line" ProcessName="DEM-FEM-Wall-Process" 
help="Assign kinematic conditions for each group containing finite elements" VariableName="WALL" 
Groups="GetPartsGroupsList">
```
Just add `Groups="GetPartsGroupsList"`

- file `kratos.gid/apps/DEM/xml/Procs.spd` add:
```xml
<proc n='GetPartsGroupsList' args='args'>
     <![CDATA[
    return [DEM::xml::ProcGetPartsGroupsList $domNode $args]
    ]]>
</proc>
```
- file `kratos.gid/apps/DEM/xml/XmlController.tcl` add:
```tcl
proc ::DEM::xml::ProcGetPartsGroupsList {domNode args} {
    return [list a b]
}
```

This will change the condition and allows us only to pick from existing groups

image
image

@maceligueta @farrufat-cimne

### DIFF
--- a/kratos.gid/scripts/Controllers/TreeInjections.tcl
+++ b/kratos.gid/scripts/Controllers/TreeInjections.tcl
@@ -234,7 +234,11 @@ proc spdAux::_injectCondsToTree {basenode cond_list {cond_type "normal"} args } 
             set state [$cnd getAttribute state]
             if {$state eq ""} {set state "CheckNodalConditionState"}
         }
-        set node "<condition n='$n' pn='$pn' ov='$etype' ovm='' icon='shells16' help='$help' state='\[$state\]' update_proc='\[OkNewCondition\]' check='$check'>"
+        set allow_group_creation ""
+        if {[$cnd getAttribute Groups] ne ""} {
+            set allow_group_creation "allow_group_creation='0' groups_list='\[[$cnd getAttribute Groups]\]'"
+        }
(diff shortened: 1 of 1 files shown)


farrufat-cimne:
thx @jginternational, i'll give it a try right away.

farrufat-cimne:
Does it work with GiD 15.0 or only in 15.1.4d?

jginternational:
> Does it work with GiD 15.0 or only in 15.1.4d?

15.1.4d or later. (Changed in GiD today, so next version will be the good one)

PR #856, opened 2021-11-18
[DEM] Change to process

jginternational:
- [ ] Check inline_vector with function [JG]


jginternational:
Just placing this here to don't forget https://github.com/KratosMultiphysics/GiDInterface/pull/854

Issue #855, opened 2021-11-16
[FluidDEM] ElementType in ProjectParameters is empty


jgonzalezusua:
Hi @jginternational ,

I am writing to you because I found two errors in the interface when it writes the ProjectParameters.json. When we introduce a particles injector the "ElementType" variable is empty but it should appear SphericPartDEMElement3D. Now appears what follows:
`"ElementType"                    : [],`

The other error is the type of particles in the Inlet.mdpa, by default GiDInterface writes:

```
INJECTOR_ELEMENT_TYPE SphericParticle3D
ELEMENT_TYPE SphericParticle3D
```

but it must write:

```
INJECTOR_ELEMENT_TYPE SphericSwimmingParticle3D
ELEMENT_TYPE SphericSwimmingParticle3D
```

Thanks for your attention

jgonzalezusua:
Hi @jginternational, could you help us to solve the issues?

jgonzalezusua:
Ok, fine.

PR #857, opened 2021-11-24
Fix FluidDem #855

jginternational:


jginternational:
You can try this branch @jgonzalezusua

jgonzalezusua:
Hi @jginternational, I am still getting an error. I do not know whether it is related to the developments of this branch . The case I use as a benchmark is this:
https://we.tl/t-CYZrg1uKAx

The error that appears is the following:
`Problem Writing MDPA block:
key "properties" not known in dictionary
End problems
Problem Writing Project Parameters block:
invalid command name ""
Event FluidDEM::write::writeParametersEvent 
End problems
`
Thanks for your attention.

jgonzalezusua:
Ok, now it works.

Thanks you very much

PR #858, opened 2021-11-29
[All] Unification of MainKratos.py

rubenzorrilla:
As discussed in [here](https://github.com/KratosMultiphysics/Kratos/issues/9341) this adds the `analysis_stage` field to the `ProjectParameters.json` input file (e.g. `"analysis_stage" : "KratosMultiphysics.FluidDynamicsApplication.fluid_dynamics_analysis"`.

This makes possible to do a dynamic import of it, that turns into the unification of all the `MainKratos.py` scripts of the interface so we will have a unique `AnalysisStageWithFlush`, which will be stored in the `exec` folder of the interface. Note that this requires to assume that the Python module name (in snake_case) matches the class name (in UpperCamelCase) (e.g. `KratosMultiphysics.FluidDynamicsApplication.fluid_dynamics_analysis` implements the class `FluidDynamicsAnalysis`). This was the case in all the cases except the `FSIApplication` (see [this PR](https://github.com/KratosMultiphysics/Kratos/pull/9410)) and the DEM application.

Besides this, the `analysis_stage` could be used for other (e.g. the unification of the ROM scripts @Rbravo555 @SADPR ) and documentation purposes.

I updated all the applications of the interface. Some comments and missing TODOs would be:
- ShallowWater: although it didn't use the analysis stage with flush I updated it accordingly to do so. @miguelmaso 
- PFEMThermic: although it didn't use the analysis stage with flush I updated it accordingly to do so. @rlrangel 
- DEM: the structure of the input is way different to that of the rest of apps so I preferred to not touch as I'm not familiar with it. I think it would be better to introduce these changes in a future PR @GuillermoCasas  @maceligueta 

I tested all the updated applications. However for the next applications I didn't run the predefined examples completely as I didn't compiled all the required apps
- ShalowWater @miguelmaso 
- PFEMFluid @rlrangel 
- PFEMThermic @AFranci 
- MPM @antonialarese @VeronikaSinger 
I note however that the input files are properly generated, meaning that the execution is interrupted (as expected) when the dynamic import takes place. Hence, I dare say that everything should be fine (letting you know just in case)


jginternational:
I'll approve this merge AFTER the release 9.0 of the GUI, and in a PR delete the json entry where the apps inform of their main script, since i'll be the same for everyone.

rubenzorrilla:
> I'll approve this merge AFTER the release 9.0 of the GUI, and in a PR delete the json entry where the apps inform of their main script, since i'll be the same for everyone.

Not sure at all, but maybe it's not a bad idea to keep that json entry (just in case). Up to you.

GuillermoCasas:
Ok, I will have a look at it with @maceligueta .

PR #865, opened 2022-02-03
Gid output folder

jginternational:
This field allows apps to store the postprocess files into a folder easily

VTK output stores the files in vtk_output, so they don't disturb the model files

This PR adds a field in the GUI
image

So all the postprocess files will be created inside a folder (except the .post.lst one, that must be on the root)

image

The effect on the ProjectParameters.json section is that the output_name has the folder prepended:
```json
"gid_output" : [{
            "python_module" : "gid_output_process",
            "kratos_module" : "KratosMultiphysics",
            "process_name"  : "GiDOutputProcess",
            "help"          : "This process writes postprocessing files for GiD",
            "Parameters"    : {
                "model_part_name"        : "FluidModelPart.fluid_computational_model_part",
                *"output_name"            : "gid_output/folder",*
                "postprocess_parameters" : {
                    "result_file_configuration" : {
                        "gidpost_flags"               : {
                            "GiDPostMode"           : "GiD_PostBinary",
                            "WriteDeformedMeshFlag" : "WriteDeformed",
                            "WriteConditionsFlag"   : "WriteConditions",
                            "MultiFileFlag"         : "MultipleFiles"
                        },
                        "file_label"                  : "time",
                        "output_control_type"         : "step",
                        "output_interval"             : 1,
                        "body_output"                 : true,
                        "node_output"                 : false,
                        "skin_output"                 : false,
                        "plane_output"                : [],
                        "nodal_results"               : ["VELOCITY","PRESSURE","TEMPERATURE"],
                        "gauss_point_results"         : [],
                        "nodal_nonhistorical_results" : []
                    },
                    "point_data_configuration"  : []
                }
            }
        }]
```

I hope this works also in MPI

### DIFF
--- a/kratos.gid/apps/Common/xml/Results.spd
+++ b/kratos.gid/apps/Common/xml/Results.spd
@@ -14,6 +14,7 @@
             <dependencies value="Yes" node="../container" att1="state" v1="normal"/>
         </value>
         <container n="GiDOptions" pn="Options" un="GiDOptions" help="GiD postprocess options" open_window="1" icon="options">
+            <value n="FolderName" pn="Folder name" v="gid_output" help="This folder will be created to store the GiD results" />
             <value n="FileLabel" pn="File Label" v="step" values="time,step" dict="time,Time,step,Step" help=""/>
             <value n="OutputControlType" pn="Units used for output frequency" v="step" values="time,step" dict="time,Time (s),step,Steps" help="" update_proc="spdAux::RequestRefresh">
                 <dependencies node="../value" actualize="1"/>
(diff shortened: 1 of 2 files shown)


rubenzorrilla:
How is the folder created? If it is created at runtime I'm afraid it won't work in MPI.

roigcarlo:
@rubenzorrilla I think this just adds the option for json to have that field. I remember reading a PR from @philbucher which added support for the creation of such folder.

There is a mechanism that hopefully prevents MPI from trying to create the folder concurrently, but I have never tested up to which point.

philbucher:
> @rubenzorrilla I think this just adds the option for json to have that field. I remember reading a PR from @philbucher which added support for the creation of such folder.
> 
> There is a mechanism that hopefully prevents MPI from trying to create the folder concurrently, but I have never tested up to which point.

I have tested it by now extensively with >2000 processes and it has not failed me once :D

philbucher:
The folder is created with the fct I added right?
Just to clarify 
Otherwise it indeed wont word properly in MPI

jginternational:
> The folder is created with the fct I added right? Just to clarify Otherwise it indeed wont word properly in MPI

I think they are created here https://github.com/KratosMultiphysics/Kratos/blob/154676d4f1624611326fe48b679d4a8a211a9d5a/kratos/python_scripts/gid_output_process.py#L753

philbucher:
> > The folder is created with the fct I added right? Just to clarify Otherwise it indeed wont word properly in MPI
> 
> I think they are created here https://github.com/KratosMultiphysics/Kratos/blob/154676d4f1624611326fe48b679d4a8a211a9d5a/kratos/python_scripts/gid_output_process.py#L753

ok then it works :+1:

Issue #873, opened 2022-04-12
[PfemFluid] Wave monitoring process


AFranci:
hi @jginternational 

I am trying to add to the PFEM-Fluid interface a process to monitor the wave height. 

I have done a first implementation and I obtained something close to what we need and I would need your help to conclude it. 
In particular, I need to put inside "[" "]" a list and give the possibility to add more points from the interface in the coordinates button (up to now, only one node can be given).

At the current point I can write from the user input the following:

    "output_processes"     : {
        "output_list" : {
            "python_module" : "wave_height_output_process",
            "kratos_module" : "KratosMultiphysics.PfemFluidDynamicsApplication",
            "Parameters"    : {
                "model_part_name"           : "PfemFluidModelPart",
                "coordinates"               : [0.0,0.0,0.0],
                "wave_calculation_settings" : {
                    "mean_water_level"       : 0,
                    "relative_search_radius" : 2
                },
                "output_file_settings"      : {
                    "file_name"   : "pfem_<i>",
                    "output_path" : "gauges"
                },
                "time_between_outputs"      : 0.01
            }
        }
    }

While I would need:
```
    "output_processes"     : {
        "output_list" :  **[**
{
            "python_module" : "wave_height_output_process",
            "kratos_module" : "KratosMultiphysics.PfemFluidDynamicsApplication",
            "Parameters"    : {
                "model_part_name"           : "PfemFluidModelPart",
                "coordinates"               : **[**
[0.0,0.0,0.0]
**..here more coordinates provided by the user...**
**]**,
                "wave_calculation_settings" : {
                    "mean_water_level"       : 0,
                    "relative_search_radius" : 2
                },
                "output_file_settings"      : {
                    "file_name"   : "pfem_<i>",
                    "output_path" : "gauges"
                },
                "time_between_outputs"      : 0.01
            }
        }
**]**
    }
```
The implementation is done in PfemFlui/ProcessForWaveMonitoring. 

Can you help me with this?

Tell me if you need more details

Regards,

Alessandro

AFranci:
Hola Javi, como lo ves?
Gracias 
Alessandro

jginternational:
Ciao Alessandro

Tras mirar con calma e intentar resolverlo, he llegado a la conclusion de que igual hay que cambiar un par de cosas.
[...]

PR #877, opened 2022-05-10
Pfem fluid/process for wave monitoring

jginternational:
Implements #873

### DIFF
--- a/kratos.gid/apps/PfemFluid/write/write.tcl
+++ b/kratos.gid/apps/PfemFluid/write/write.tcl
@@ -1,7 +1,7 @@
 namespace eval ::PfemFluid::write {
     namespace path ::PfemFluid
     Kratos::AddNamespace [namespace current]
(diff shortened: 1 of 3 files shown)


AFranci:
Hi @javi 
Thank you again for your help. I made some changes...
The current version is working correctly. Before, the analyses were not running because there were two parentheses [[ ]] for the  "output_list". I fixed it but I will tell me if it's the best way...
I like the idea of selecting the points, I agree that it can be very easy to use. However, I don't need to create a submodel part for this because I will not use it during the computation. That's why I removed it from the mdpa. As before, please check if it looks ok for you.
Finally, I have added the option of giving the name of the folder collecting wave measures of the same group. This can be very useful when you need to monitor the wave along different lines in three-dimensional problems.

AFranci (approved):
For me it's ok to merge it as it is

PR #891, opened 2022-08-31
[Structural] Adding "use_old_stiffness_in_first_iteration" as true by default

AlejandroCornejo:
@RiccardoRossi @rubenzorrilla @loumalouomega 

As we know, in non-linear simulations in solids is mandatory to add always the "`use_old_stiffness_in_first_iteration`"  to be true in the json in order to properly work.

In this PR I am adding this line an setting it to true always. De we agree on this?

Regards

### DIFF
--- a/kratos.gid/apps/Structural/write/writeProjectParameters.tcl
+++ b/kratos.gid/apps/Structural/write/writeProjectParameters.tcl
@@ -90,6 +90,8 @@ proc ::Structural::write::getOldParametersDict { } {
     dict set solverSettingsDict material_import_settings $materialsDict
 
     # Solution strategy parameters and Solvers
+    # We now added this variable as true by default (in solids is mandatory)
+    dict set solverSettingsDict use_old_stiffness_in_first_iteration true
     set solverSettingsDict [dict merge $solverSettingsDict [write::getSolutionStrategyParametersDict STSolStrat STScheme STStratParams] ]
     set solverSettingsDict [dict merge $solverSettingsDict [write::getSolversParametersDict Structural] ]
 


rubenzorrilla (changes requested):
By doing it in this way you'll be setting it to `true` also in linear simulations. Besides, you're not exposing it to the GUI.

Instead, I'd expose this parameter in the GUI, making it active if a non-linear analysis is selected. You can default it to true in there.

AlejandroCornejo:
Now the parameter is set only in non_linear simulations. IMO, in non linear constitutive problems always should be true so I would not expose it to the GUI. If we need to expose it I'd need the help of  @jginternational  to do that :)

rubenzorrilla:
> Now the parameter is set only in non_linear simulations. IMO, in non linear constitutive problems always should be true so I would not expose it to the GUI. If we need to expose it I'd need the help of @jginternational to do that :)

I think it must be exposed. Note that it has an extra computational cost, which might be not always required. You can default it to `true`. I don't think it is a good idea to hardcode it like this. If you want to hardcode things, the GiD GUI scripts is not the place as you will be only hardcoding for the GiD users. The place is the solver defaults.

jginternational:
I'd recommend the changes above so the parameter is shown in gui

AlejandroCornejo:
Now it is exposed to the GUI but set false by default.

AlejandroCornejo:
I think now should be fine right @rubenzorrilla ?

jginternational (approved):

rubenzorrilla:
> I think now should be fine right @rubenzorrilla ?

It is. Thanks!

PR #894, opened 2022-09-01
Dam/nodal young modulus

jginternational:


joaquinirazabal:
Hi, @jginternational. I have been checking this branch and I think that we can delete it.

PR #895, opened 2022-09-01
restart options

jginternational:


maceligueta:
The restart feature is not fully operative for the DEM. It just works for very simple cases. I would not add the capability to DEM. @GuillermoCasas can confirm this, I guess.

PR #901, opened 2022-09-21
Copy custom GeoMechanicsApplication problem type from Kratos repo to GiDInterface

carloslubbers:
Related to https://github.com/KratosMultiphysics/Kratos/pull/9956


rubenzorrilla:
@jginternational can you confirm this is OK?

jginternational (changes requested):
This is a problemtype itself, the old way the problemtypes were done.

I can help you with the migration to the apps system inside the kratos problemtype.
Check how other apps are done, how they define the tree structure, the elements, the conditions, processes... I think this app can be similar to the Structural or Convection diffusion one.

PR #906, opened 2022-10-25
Refining box can be repeated

jginternational:
image


AFranci:
Many thanks @jginternational it's perfect!

PR #907, opened 2022-11-12
Mpm/updates

lauramoremar:
**📝 Description**
Some updates in the interface of the MPM application.

**🆕 Changelog**
- The Model Information section has been extended.
- The "Solid" domain is renamed as "Material_domain" in the "Parts" section.
- The UP formulation can be chosen as element, but for this moment only the hyperelastic material is eligible for this.
- The gravity has now a "On/Off" bottom (by default On option is activated).
- Time parameters are all inside the "Strategy Parameters", and the "Time intervals" section has been deleted (because, until I know, it is not implemented in the MPM code).
-  Some useless options inside "strategy parameters" have been removed.
- The Vtk postprocess option can be selected from the gid interface now and it works perfectly.

This image clarify the main changes:
GiD_interface_MPM_updates


lauramoremar:
@VeronikaSinger could you check it before the merge? 
We are checked it in ubuntu, but I think it would be interesting to check it also in Windows :)

kratos.gid/apps/MPM/write/write.tcl
```diff
@@ -90,6 +90,31 @@ proc MPM::write::GetPartsGroups { part_type {what "name"} } {
...
+        set elem_name [write::getValueByNode [$gNode selectNodes ".//value\[@n='Element']"] ]
+        set e [Model::getElement $elem_name]
+        if {$get eq "Name"} { set e [$e getName] }
+        lappend lista $e
+    }
+    return $lista
+}
+
+
+proc ::MPM::write::GetPartsGroups2 { {get "Objects"} } {
```
jginternational, line 106: I think that this function is the same as -> write::getPartsGroupsId
https://github.com/KratosMultiphysics/GiDInterface/blob/21ef2c92d01b63ee68d2aa793b1fc715d2661eaf/kratos.gid/scripts/Writing/Writing.tcl#L413

As the one in writing is a common code, I think you can refer to it.

jginternational (changes requested):

kratos.gid/apps/MPM/write/write.tcl
lauramoremar: You are completely right. I've changed it in the new commit. Thank you! :)

VeronikaSinger:
Hi @lauramoremar 

thanks for the updates! It works perfectly for me as well :)
Possibly you could rename the Material_domain without the underscore to be consistent with the other names.

lauramoremar:
Hi @VeronikaSinger , yes I agree with you!

I was finding out about modify it, but the problem is that the name requires "non-white spaces" inside it,  due to we are calling the "injectPartsByElementType" in "MPM/Parts.spd". Here the internal name is automatically inserted as the name in the tree (and the internal names are always without white spaces).
The injectPartsByElementType is a "common procedure" and  I'm not sure if I can/should modify it. I guess I could create another procedure inside MPM to take into account that, but I don't know what is the best way...

@jginternational what do you recommend us in that case?

jginternational:
@lauramoremar 
> 
> @jginternational what do you recommend us in that case?

I think this ([d6984db](https://github.com/KratosMultiphysics/GiDInterface/pull/907/commits/d6984db98d91d4293174c28364b3605e7954ae0e)) answers the question
It never happened before, but it's a good point to replace _ by " "

👍

jginternational (approved):

lauramoremar:
> @lauramoremar
> 
> > @jginternational what do you recommend us in that case?
> 
> I think this ([d6984db](https://github.com/KratosMultiphysics/GiDInterface/pull/907/commits/d6984db98d91d4293174c28364b3605e7954ae0e)) answers the question It never happened before, but it's a good point to replace _ by " "
> 
> +1

Yes of course! 
Thank you @jginternational ! :)

Issue #909, opened 2022-11-24
[MPM] Visibility of a strategy parameter which depends on the element type


lauramoremar:
Hi, 
I would like to make visible one option in the "Strategy parameters" section but only if I'm using a mixed Element.

I tried to do it using the **getStateFromXPathValue** procedure, but it didn't work. Probably I was doing something wrong:

`<value n="ActivateStabilization" pn="Stabilization" un="MPStabilization" help="Stabilization" open_window="0" actualize_tree="1" state="[getStateFromXPathValue {string(container[@n='MPM']/container[@n='Parts']/condition[@n=Parts_Material_domain]/value[@n=Element]/@v)} UpdatedLagrangianUP2D]]" />`

I tried to use the **checkStateByUniqueName** introducing in the container element a unique name (un=MPMElement) in the procedure injectPartsByElementType (when the container is built for each element), but it doesn't work either.

Any ideas on how to do it? 
Thank you! :)

PR #912, opened 2022-12-21
[MPM] Examples and Newtonian fluid as constitutive law

lauramoremar:
**📝 Description**
- Newtonian fluid constitutive law for irreducible element added.
- Stabilization option on/off. (it only works for the mixed formulation) _It would be nice if the stabilization option only was visible when the mixed formulation is employed, but for the moment I were incapable of doing that ([Issue 909](https://github.com/KratosMultiphysics/GiDInterface/issues/909)), and it is always visible._
- New examples (Static and dynamic cantilever beam and granular flow)
- Loads are now inside on Boundary Conditions section
- Gravity is inside on Solution section (as in the Structural interface)

I attach two pictures for clarifying the changes:
<p align="center">
  
  
</p>


lauramoremar:
In the files changed in this commit you will see some changes already done in a previous pull request. This is because in this commit [3d88ec97](https://github.com/KratosMultiphysics/GiDInterface/commit/3d88ec97ba445523754f3ed9cf7fe929b865ed0d) , I don't know why, a "revert" of these changes have been done. :S

lauramoremar:
if none of you have any comments and you are agree with the changes I will do the merge tomorrow :)

PR #921, opened 2023-03-03
Grid mdpa should contain only grid nodes

### LINKED ISSUE #920: [MPM] Duplicated points in mdpa file
Hi,

In the Particle Mechanics Application we need to define two computational meshes, the **background** grid and the **body** grid.

When exporting the data, we obtain two `*.mdpa` files, each of them containing the information of one of the two meshes, say `problem_Grid.mdpa` and `problem_Body.mdpa`.

We have noticed that in the in the file `problem_Grid.mdpa` there are all the nodes, i.e. the ones defining the **background** grid and the ones defining the **body** grid. 

However we would like to have only the nodes defining the background grid in the file `problem_Grid.mdpa`

In the images below there is an example. On the left there is part of the file `problem_Body.mdpa`, on the right part of the file `problem_Grid.mdpa`. With the red circle I denoted the nodes of the **body** mesh that are repeated on both files.

How can we remove the nodes used for the body mesh from the background mesh file?

Thanks in advance.

issue

beam

jginternational:
Fix #920 

image

Hope this helps

### DIFF
--- a/kratos.gid/apps/MPM/write/write.tcl
+++ b/kratos.gid/apps/MPM/write/write.tcl
@@ -38,7 +38,7 @@ proc MPM::write::writeModelPartEvent { } {
     write::WriteString "End Properties"
 
     # Nodal coordinates
-    write::writeNodalCoordinates
+    write::writeNodalCoordinatesOnGroups [MPM::write::GetPartsGroups grid]
 
     # Grid element connectivities
     writeGridConnectivities


lauramoremar:
Thank you Javi!!!!

ncrescenzio:
After this change, all the nodes of a sub model part of the grid are not included in the list of nodes at the beginning of the file `*_Grid.mdpa`. For example, in the file `example_Grid.mdpa`, all the nodes included in the section
```
Begin SubModelPart Slip2D_Slip_Auto1 // Group Slip Auto1 // Subtree Slip2D
   Begin SubModelPartNodes
          42
          58
          75
          ...
   End SubModelPartNodes
   ...
End SubModelPart
```
are missing in the section `Begin Nodes ... End Nodes` at the beginning of the file.

How can we include this nodes?

[example.gid.zip](https://github.com/KratosMultiphysics/GiDInterface/files/13042748/example.gid.zip)

jginternational:
This happens because in this mode of writing the nodes, we are printing in each mdpa the nodes assigned to Parts
In Grid.mdpa the nodes that belong to groups assigned to the Part Grid
In Body.mdpa the nodes that belong to groups assigned to the Part Body

In this case, I can see that the model part Slip has a line that do not belong to any of those parts, so it's nodes are not being printed.

In 2D I can see 4 conditions available:
- Loads on points
- Loads on lines
- Slip
- Dispacement

I'm going to make a PR with a fix that writes the nodes of the conditions in the Grid.mdpa

PR #922, opened 2023-03-03
Hotfix mpm stabilization

### LINKED ISSUE #909: [MPM] Visibility of a strategy parameter which depends on the element type

jginternational:
Fix #909 

image

image

@lauramoremar Can you check it?

### DIFF
--- a/kratos.gid/apps/MPM/write/write.tcl
+++ b/kratos.gid/apps/MPM/write/write.tcl
@@ -90,10 +90,9 @@ proc MPM::write::GetPartsGroups { part_type {what "name"} } {
     return $body_groups
 }
 
-
 proc ::MPM::write::GetUsedElements { {get "Objects"} } {
     set lista [list ]
-    foreach gNode [write::getPartsGroupsId node] {
+    foreach gNode [MPM::write::GetPartsGroups Body node] {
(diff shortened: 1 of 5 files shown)


lauramoremar:
Hi Javi!! That's exactly what I wanted to do.
Did you delete the branch by mistake or did you see something wrong? :S

jginternational:
Something strange has hapened with my Github.
I accidentaly closed some branches

lauramoremar:
I have checked it and it works perfectly. 
Thank you!!! :)

PR #944, opened 2023-07-06
Get format dict for geometries no mid

jginternational:
Remove properties id in mdpa


rubenzorrilla:
Are you sure this is working? I'm getting this error `expected integer but got ""` when using the MDPA generator.

jginternational:
> Are you sure this is working? I'm getting this error `expected integer but got ""` when using the MDPA generator.

Wow, I can't reproduce it @rubenzorrilla

rubenzorrilla:
I got it with the latest developer version and current master branch of the GUI (in Linux environment).

PR #951, opened 2023-09-15
Fluid write with geometries

jginternational:
This PR implements the geometry write just for FLUID app, and the project parameters with the modelers and so.
All apps depending on Fluid will still write elements and conditions

To be merged after https://github.com/KratosMultiphysics/Kratos/pull/11370


jginternational:
This PR implements the geometry write just for FLUID app
All apps depending on Fluid will still write elements and conditions

rubenzorrilla:
Posting as a reminder that we should only use this in the serial case (for the moment until we implement the partitioning of geometries).

jginternational:
image

In the Kratos preferences window, I've enabled this entry, so users can enable the new write for those apps who implement it.
Currently Fluid and Geomechanics only.
Apps derived from Fluid will still print element and conditions

jginternational:
> Posting as a reminder that we should only use this in the serial case (for the moment until we implement the partitioning of geometries).

image

This is the message shown if the user selects MPI and write geometries, wich is not allowed

PR #960, opened 2023-10-18
[Geomechanics] Add conditions and improve example

jginternational:
Added some conditions and nodal conditions extracted from the example provided by Deltares team

Created an example that uses those conditions
Improved the code for writing geometries and moving some code from Fluid app to common


avdg81 (approved):
Well done! This looks really nice, so let's merge it :-)

PR #971, opened 2024-04-30
[Fluid] Div-stable element P2-P1

rubenzorrilla:
Adding div-stable P2-P1 element.


jginternational (approved):

rubenzorrilla:
Lets wait for the merge until it is also available in Kratos master.

PR #979, opened 2024-06-03
[Fluid] Pressure conditions

rubenzorrilla:
This adds the possibility to set custom pressure conditions similarly to the custom velocity constraints. So far I only made it available for points in 2D and points and lines in 3D.

@jginternational we need to confirm that conditions are never written for this. In other words, we only need to write nodes in the submodelparts to which we assign this condition.


rubenzorrilla:
> It is currently set as: 2D -> Point & Lines -> Nodes 3D -> Point, lines & surfaces -> Nodes
> 
> If it's fine for you, you can merge

Provided that only nodes are written in the corresponding submodelparts it should be fine.

PR #980, opened 2024-06-03
[Fluid] Update P2P1 example

rubenzorrilla:
As we agreed, in this PR I'm upgrading the P2P1 element example. Hence, instead of solving (again) the 100Re example we solve the lid-driven cavity flow.

Note that this requires #979 to be merged first.


jginternational (approved):
It works,

I'd probably dont fix the mesh size, in the example tcl, but I understand that it is important to get a proper RN

rubenzorrilla:
> It works,
> 
> I'd probably dont fix the mesh size, in the example tcl, but I understand that it is important to get a proper RN

Well, we can leave it free, but if so you can get whatever... Considering that in the other examples we provide some basic mesh settings I do so here too.

PR #985, opened 2024-07-11
MIGRATION - Write as geometries

jginternational:
This PR / branch will accumulate the migration of the different apps from the classic MDPA write to the new format as geometries

- [x] Fluid
- [x] Embedded Fluid #989 
- [x] Geomechanics
- [x] Compressible flow #991 
- [x] Convection diffussion #992 
- [x] Structural #988 
- [x] Buoyancy #986 
- [x] Shallow water #995  
- [x] Free surface #990 ➡️ Wait for https://github.com/KratosMultiphysics/Kratos/issues/12564
- [x] Conjugate heat transfer #994 
- [x] FSI #996 ➡️ Wait for https://github.com/KratosMultiphysics/Kratos/issues/12747
- [x] Potential Flow #987 #1005
- [x] MPM #1004 
- [ ] DAM
- [ ] PFEM Fluid
- [ ] DEM
- [ ] DEM PFEM
- [ ] Fluid DEM
- [ ] PFEM Melting

PR #988, opened 2024-07-18
Migrate structural

jginternational:
Another one I can not run @rubenzorrilla 

[structure_as_geometry.gid.zip](https://github.com/user-attachments/files/16284900/structure_as_geometry.gid.zip)

### DIFF
--- a/kratos.gid/apps/Structural/app.json
+++ b/kratos.gid/apps/Structural/app.json
@@ -51,7 +51,8 @@
         "properties_location": "json",
         "model_part_name": "Structure",
         "output_model_part_name": "",
-        "enable_dynamic_substepping": true
+        "enable_dynamic_substepping": true,
+        "write_mdpa_mode": "geometries"
     },
     "main_launch_file": "../../exec/MainKratos.py",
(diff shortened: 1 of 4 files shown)


jginternational:
Depends on #985

rubenzorrilla:
[structure_as_geometry.gid.zip](https://github.com/user-attachments/files/16312209/structure_as_geometry.gid.zip)

To make it work, I added:
- properties 1 to the mdpa
- assigning the properties to the entire model part so the conditions also have properties (hacky, as I conditions should have different properties than the elements)

rubenzorrilla:
[structure_as_geometry.gid.zip](https://github.com/user-attachments/files/16312315/structure_as_geometry.gid.zip)

I'd do something like this
- there are no properties in the mdpa
- we assign the "fake" 0 properties to the conditions in the  `StructuralMaterials.json` (we can hardcode this for the moment)

rubenzorrilla:
Indeed, al these issues come from the fact that our `gid_output_process.py` maps our properties to output layers. Rather than adding fake properties to the materials json, I think we'd better amend the output stuff.

PR #990, opened 2024-07-18
Free surface writes geometries

jginternational:
Free surface fails running because of this:

In the first picture we can see the the group Fluid:
image

In the second picture we can see the group Surface bottom:
image

They do share elements

In the MDPA, both groups are written as `Begin Geometries Triangle2D3`
So there are elements declared 2 times, with the same id and the same connectivities. This should not be a problem, but it is.

You can see the log:

```

  File "C:\Users\garat\Desktop\Deploy_Kratos\Deploy_Kratos\KratosMultiphysics\analysis_stage.py", line 80, in Initialize
    self._ModelersSetupModelPart()
  File "C:\Users\garat\Desktop\Deploy_Kratos\Deploy_Kratos\KratosMultiphysics\analysis_stage.py", line 285, in _ModelersSetupModelPart
    modeler.SetupModelPart()
  File "C:\Users\garat\Desktop\Deploy_Kratos\Deploy_Kratos\KratosMultiphysics\modelers\import_mdpa_modeler.py", line 36, in SetupModelPart
    KratosMultiphysics.SingleImportModelPart.Import(
RuntimeError: Error: Attempting to add a new geometry with Id :892, unfortunately a (different) element with the same Id already exists




in kratos/includes/model_part.h:1538: ModelPart::AddGeometries
   kratos/includes/model_part.h:1560: ModelPart::AddGeometries
   kratos/sources/model_part_io.cpp:1771: ModelPartIO::ReadGeometriesBlock
   kratos/sources/model_part_io.cpp:618: ModelPartIO::ReadModelPart


```

Then in the submodelparts definition, each group references it's triangles.

### DIFF
--- a/kratos.gid/apps/FreeSurface/app.json
+++ b/kratos.gid/apps/FreeSurface/app.json
@@ -38,7 +38,7 @@
         "properties_location": "json",
         "model_part_name": "FluidModelPart",
         "output_model_part_name": "",
-        "write_mdpa_mode": "entities"
+        "write_mdpa_mode": "geometries"
     },
     "main_launch_file": "../../exec/MainKratos.py"
 }
--- a/kratos.gid/apps/FreeSurface/write/writeProjectParameters.tcl
+++ b/kratos.gid/apps/FreeSurface/write/writeProjectParameters.tcl
@@ -25,6 +25,9 @@ proc ::FreeSurface::write::getParametersDict { } {
 
     set projectParametersDict [::write::GetModelersDict $projectParametersDict]
 
+    set projectParametersDict [::FreeSurface::write::ReplaceConditionsName $projectParametersDict "WallCondition2D2N" "LineCondition2D2N"]
+    set projectParametersDict [::FreeSurface::write::ReplaceConditionsName $projectParametersDict "WallCondition3D3N" "LineCondition3D3N"]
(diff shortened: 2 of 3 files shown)


jginternational:
Example attached
[freesurface_geom.gid.zip](https://github.com/user-attachments/files/16312550/freesurface_geom.gid.zip)

rubenzorrilla:
After having discussed this with @jginternational we concluded that the issue here is having something similar to
```
Begin Geometries Triangle2D3
    1 10 20 30
    1 10 20 30
    2 20 30 40
    ....
End Geometries
```
that is having the same entity repeated. Note that this is to happen in many cases (for instance having a shell with surface load conditions).
Right now, we throw an error if the same geometry id is present, but the error should in our opinion be thrown if the **same id with different connectivities is present**. If the **same id with same connectivities** is only repeated information which should be simply skipped. @KratosMultiphysics/technical-committee

rubenzorrilla:
Note that checking for the repeated information in the writing of the pre (regardless this is GiD or another one) leads to custom and potentially inefficient writing implementations.

rubenzorrilla:
Should be fixed by https://github.com/KratosMultiphysics/Kratos/pull/12703.

PR #991, opened 2024-07-18
Compressible migration geoms

jginternational:
Not working.
It gets stacked

[compres_new.gid.zip](https://github.com/user-attachments/files/16286484/compres_new.gid.zip)


jginternational:
Working!

PR #992, opened 2024-07-18
convection diffussion writes as geoms

jginternational:
It gets stacked
[convdiff_new.gid.zip](https://github.com/user-attachments/files/16286574/convdiff_new.gid.zip)


jginternational:
Ready now!

PR #994, opened 2024-10-09
Conjugate heat transfer migration as geometries

jginternational:


jginternational:
@rubenzorrilla we need to check this one.
It writes input files but I think that calculation crashes, so probably there is something wrong there.
[cht3.gid.zip](https://github.com/user-attachments/files/17334267/cht3.gid.zip)

rubenzorrilla:
This one should work with the changes I'm about to push to the Kratos repo [cht3_RZ.zip](https://github.com/user-attachments/files/18026398/cht3_RZ.zip)

jginternational:
solo queda revisar este modeler
image

jginternational:
@rubenzorrilla please check this case

[aaa.gid.zip](https://github.com/user-attachments/files/18537706/aaa.gid.zip)

rubenzorrilla:
> @rubenzorrilla please check this case
> 
> [aaa.gid.zip](https://github.com/user-attachments/files/18537706/aaa.gid.zip)

As far as I see, the section
```json
            },{
                "model_part_name" : "FluidThermalModelPart.Fluid_Left_Wall",
                "condition_name"  : "ThermalFace2D2N"
```
in the `CreateEntitiesFromGeometriesModeler` shouldn't be there. Aside of this I don't see anything strange.

However, when I ran the case I got zero residual every time step. Is the case supposed to work or it is a `random` one just for the sake of checking the I/O?

jginternational:
@rubenzorrilla it comes from the 2nd example in the GUI

rubenzorrilla:
> @rubenzorrilla it comes from the 2nd example in the GUI

Then it should work, meaning that I missed something else. I'll try to have another look ASAP.

jginternational:
Located a problem here:
- FluidMaterials.json:
    - Contains: FluidThermalModelPart
    - Should contain: FluidModelPart.Fluid
- BuoyancyMaterials.json
    - Contains ThermalModelPart
    - Should contain FluidThermalModelPart

Check:
- Buoyancy migrated app
- Restore namespace variables after print projectparameters.json

PR #1015, opened 2025-07-24
[MPM] Compute reactions flag

jginternational:
Hi @ncrescenzio 

Please change this draft into a PR when you feel that this changes can fit into MPM applications in production


lauramoremar:
Hi! I think we should include this option in the interface, since it is useful when reactions are requested for plotting in the postprocess or for computing forces into a file. My only doubt is whether this is the correct section **Strategy parameters** in the interface, or if it should instead belong to the postprocess section, **Results** (although, at the same time, it is also something that needs to be computed...).

image

What do you think? @ncrescenzio @andimkatili @antonialarese

ncrescenzio:
@lauramoremar I would leave it here because it is more consistent with the structure of the json parameters. Moreover this option is also used to initialize the [`MPMResidualBasedNewtonRaphsonStrategy`](https://github.com/KratosMultiphysics/Kratos/blob/55ef302a7769c4ab7985c204ceccef1f1f8ba3cd/applications/MPMApplication/python_scripts/mpm_solver.py#L422-L429), so I think it makes sense to leave it here

lauramoremar (approved):

PR #1023, opened 2026-01-05
Simulations history

jginternational:
The idea is to keep a history of the simulations calculated of a case, and, in a mid term future, when GiD adapts, be able to compare results from 2 different simulations.

Every time you click on run, the problemtype creates a subfolder for the simulation, and the input files and python script are copied there.
 In the tree, user can see the history like this:
image
With the red icon if the calculation failed, and green if everything is OK

For every simulation, the user can:
- Open the case in VSCode
- Delete (will delete the results)
- Rerun
- View result (Goes to postprocess and load the results)
- View error log (only if error)
image

Pending tasks
- At this point, we need to work on the log view in real time. GiD expects the .info file to be in the model folder.
- Add an spd viewer, in 'disabled' mode, so the user can see the state of the tree that generates the run. now the only way is to Open with VSCode and view the mdpa and Project parameters (maybe it's enough)
- In future GiD versions, add "Compare results". the user will be able to compare 2 or more simulations, step by step.

This PR is open to new features and comments!


jginternational:
Added the launch window:
Features:
- [x] Set launch name 
- [x] Enable/disable gid/vtk outputs
- [x] Skip this window
- [ ] Launch location: local, cloud, hpc custom
- [ ] Launch mode if local

PR #1031, opened 2026-05-19
[MPM] Fixing gravity bug

lauramoremar:
There was a bug that did not allow users to modify the gravity direction and modulus from the GiD interface.

This has been fixed by removing the rules that overwrote the MPM gravity values whenever gravity was enabled. 
The default gravity values are now defined directly in the editable fields.

Screenshot from 2026-05-19 15-27-55


ncrescenzio (approved):
Thanks @lauramoremar!

PR #1034, opened 2026-05-21
[MPM] Improvement "Results -> On nodes" output fields configuration

lauramoremar:
### Changes

- Adds missing nodal result fields in `Results > On nodes`:
  - `ACCELERATION` 
  - `VELOCITY`
  - `PRESSURE`

- Shows and enables `PRESSURE` by default only when a mixed UP  formulation is used.
- The pressure result options are hidden for non-mixed formulations to avoid exposing unavailable output fields.
- Applies the same mixed-formulation visibility logic to `MP_PRESSURE` in `Results > On element`.

image

I would appreciate it if you could try this locally, in case I missed anything. Suggestions for improvement are also very welcome.

### DIFF
--- a/kratos.gid/apps/MPM/xml/Elements.xml
+++ b/kratos.gid/apps/MPM/xml/Elements.xml
@@ -23,9 +23,9 @@
       <parameter n="MATERIAL_POINTS_PER_ELEMENT" pn="Material points per element" type="combo" v="1" values="1,3,4,6,9,12,16,25" pvalues="1 (for triangles and quads),3 (for triangles),4 (for triangles and quads),6 (for triangles),9 (for quads),12 (for triangles),16 (for quads),25 (for quads)" help="Number of material points generated for each element. Available options are the following. Triangular: 1,3,4,6,12 and Quadrilateral: 1,4,9,16,25" />
     </inputs>
     <outputs>
+      <parameter n="MP_ACCELERATION" pn="Material point acceleration" v="false" />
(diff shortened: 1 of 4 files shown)


ncrescenzio (commented):
I tested the feature you are proposing and it works on my computer.

Just a small request... can you please put the variables in the same order in the two sections? 

Image

lauramoremar:
> I tested the feature you are proposing and it works on my computer.
> 
> Just a small request... can you please put the variables in the same order in the two sections?
> Image

When I modified it, I thought the same thing. But it is not easy to change the order of the fields in that section due to the structure of the file:
[kratos.gid/apps/MPM/xml/NodalConditions.xml](https://github.com/KratosMultiphysics/GiDInterface/pull/1034/changes#diff-1e0bb9f3c22f4ec2df7f00b39adaf1708c821295b0931850bc7ce65f08d57675).

It seems that Reaction and Pressure internally "depend" on Displacement.
I’m not sure how we could modify that. Any ideas, @jginternational?

Another option could be trying to change the order of the fields "on element", which seems easier to implement.

lauramoremar:
I modified  [Elements.xml ](https://github.com/KratosMultiphysics/GiDInterface/pull/1034/changes#diff-14e03e68d499e13fda971702be2548f4698d693188c3d6beaa649f211f2eaf94):

image

Now, both sections follow the same order.

ncrescenzio (approved):
Thanks @lauramoremar

